PR #3, opened 2020-04-22
First round of changes

### LINKED ISSUE #1: GPL-407 Store sample data from external sites
**User story**
As a dev team we would like a service to store sample data from external sites. This data arrives as CSV files on an internal SFTP server.

**Who the primary contacts are for this story**
Philip

**Acceptance criteria**
1. connect to and download the latest CSV files for a site
2. update (drop and recreate) the sample data for that site

**Additional context**

pjvv:
Closes #1 

* Add everything required to store samples to a mongodb instance


crawler/main.py
```diff
@@ -0,0 +1,68 @@
...
+            imports_collection = get_mongo_collection(db, COLLECTION_IMPORTS)
+            samples_collection = get_mongo_collection(db, COLLECTION_SAMPLES)
+
+            if samples_collection.estimated_document_count() > 0:
+                logger.info(f"{COLLECTION_SAMPLES} collection is not empty so creating copy")
+                copy_collection(db, samples_collection)
+
+                logger.info(f"Removing all documents from {COLLECTION_SAMPLES}")
+                result = samples_collection.delete_many({})
+                logger.debug(f"{result.deleted_count} records deleted")
```
emrojo, line 40: Why do we archive the old samples in a different collection by timestamp? Is not something that could be equivalent to keeping them with their import date timestamp?

crawler/main.py
```diff
@@ -0,0 +1,68 @@
...
+
+                    errors, docs_to_insert = parse_csv(centre)
+
+                    logger.debug(f"Attempting to insert {len(docs_to_insert)} docs")
+                    result = samples_collection.insert_many(docs_to_insert)
+                    logger.debug(f"{len(result.inserted_ids)} documents inserted")
+
+                    # write status record
+                    # TODO create the status record somewhere that if something else fails it is
+                    # still written
```
emrojo, line 57: Is this about an external service? 
Why can't  we store a status of 'processing' before import, and then a status 'finished' after to keep track of this possible failure?

crawler/main.py
emrojo: Ok, when we know better what is going to be in the inputs we can decide

pjvv: Need to double check with Rich.

PR #5, opened 2020-05-05
Add support to use MONGO_URI

emrojo:


crawler/db.py
```diff
@@ -20,21 +20,28 @@ def create_mongo_client(config: Dict[str, str]) -> MongoClient:
...
-        host=mongo_host,
-        port=mongo_port,
-        username=mongo_user,
-        password=mongo_password,
-        authSource=mongo_db,
-    )
+
+    if "MONGO_URI" in config:
+        mongo_uri = config["MONGO_URI"]
+        logger.info(f"Connecting to {mongo_uri}")
```
pjvv, line 26: Doesn't this show the password too?

PR #12, opened 2020-05-13
Gpl439 merge daily files

### LINKED ISSUE #4: GPL-439	Merge daily files into a single file and add this to daily job
**User story**
GPL-439 | As a LIMS manager (Rich) I may want what we think are incremental daily updates from UK Biocenter aggregated into a single file to a) Make it work without further changes to the automated lighthouse code b) provide the file for the SSR to manage the sample picking strategy (esp, Siobhan)

**Who the primary contacts are for this story**
Rich L
Siobhan

**Acceptance criteria**
I as thinking... 

1. cmd line app 
2. Input
2.1. collection lighthouse plate map files
2.2. a Lab id e.g. MK
3. Output
3.1. Single file agregating unique samples by plate + 1 header row
3.2. log of 
4. Samples lines iwth unexpected format
5. Assume if plate is on more than one file thi is a safe fail.
6. any of team could run (Needs readme)

**Additional context**

pjvv:
Closes #4


crawler/helpers.py
```diff
@@ -97,41 +109,65 @@ def check_for_required_fields(csvreader: DictReader, centre: Dict[str, str]) ->
     Returns:
         List[str] -- list of errors experienced
     """
-    logger.debug("Checking data for required fields")
-
-    if csvreader.fieldnames and centre["barcode_field"] not in csvreader.fieldnames:
-        error = "Barcode field not in CSV file"
-        logger.error(error)
-        return [error]
+    """Checks that the CSV file has the required headers.
```
JamesGlover, line 112: Might be a pythonism I'm missing, but is there a reason this method has two docstrings? Is it a way of indicating that the method will either Return OR Raise?

JamesGlover: Actually, is the former docstring still accurate? We don't seem to return a list of errors any more.

crawler/helpers.py
```diff
@@ -97,41 +109,65 @@ def check_for_required_fields(csvreader: DictReader, centre: Dict[str, str]) ->
...
+    required_fields = {
+        FIELD_ROOT_SAMPLE_ID,
+        FIELD_RNA_ID,
+        FIELD_RESULT,
+        FIELD_DATE_TESTED,
+        FIELD_LAB_ID,
+    }
+    if csvreader.fieldnames:
+        fieldnames = set(csvreader.fieldnames)
+        if not required_fields < fieldnames:
```
JamesGlover, line 127: Should this be required_fields <= fieldnames? Otherwise it'll error if the two fieldsets are the same, but the actual error message then generated will be confusing. (I'm assuming it would say ' missing in CSV file'). I figure you'll almost always have additional fields anyway, so perhaps this is just a technicality.

crawler/helpers.py
```diff
@@ -144,52 +180,199 @@ def parse_csv(centre: Dict[str, str]) -> Tuple[str, List[str], List[Dict[str, st
...
-    (_, _, files) = next(os.walk(root.joinpath(DIR_DOWNLOADED_DATA)))
-
-    pattern = re.compile(centre_details["sftp_file_regex"])
-    centre_files = filter(pattern.match, files)
+    centre_files = get_files_in_download_dir(centre, regex_field)
 
     files_with_time = {}
     for filename in centre_files:
-        if match := pattern.match(filename):
+        if match := re.compile(centre[regex_field]).match(filename):
```
JamesGlover, line 258: How fast is regex compiling? Is it worth caching it outside of the loop?

crawler/helpers.py
```diff
@@ -144,52 +180,199 @@ def parse_csv(centre: Dict[str, str]) -> Tuple[str, List[str], List[Dict[str, st
...
+    #   https://bitbucket.org/dundeemt/pysftp/src/master/docs/cookbook.rst#rst-header-id5
+    cnopts = pysftp.CnOpts()
+    cnopts.hostkeys = None
+
+    sftp_host = config.SFTP_HOST  # type: ignore
+    sftp_port = config.SFTP_PORT  # type: ignore
+    sftp_user = config.SFTP_WRITE_USER  # type: ignore
+    sftp_password = config.SFTP_WRITE_PASSWORD  # type: ignore
+
+    with pysftp.Connection(
```
JamesGlover, line 336: This sftp connection code looks shared with that in download_csv_files. Can it be extracted easily?

crawler/main.py
```diff
@@ -19,71 +26,117 @@
...
+                    #   the records from being written
+                    logger.warning(e)
+                    docs_inserted = e.details["nInserted"]
+                    write_errors = {write_error["code"] for write_error in e.details["writeErrors"]}
+                    for error in write_errors:
+                        num_errors = len(
+                            list(filter(lambda x: x["code"] == error, e.details["writeErrors"]))
+                        )
+                        errors.append(f"{num_errors} records with error code {error}")
+                except Exception as e:
```
JamesGlover, line 130: As mentioned in slack, this is where I think I'd either expect to see handling of the CentreFileError exception, or possibly just populating the errors list with the more general exceptions (I'm not sure what else ends up in here). Again, may easily be missing something.

crawler/utils.py
```diff
@@ -16,16 +16,7 @@ def send_message(self, sent_str):
...
-                    {
-                        "type": "section",
-                        "text": {
-                            "type": "mrkdwn",
-                            "text": "Danny Torrence left the following review for your property:",
-                        },
-                    },
-                    {"type": "section", "text": {"type": "mrkdwn", "text": sent_str}},
-                ],
+                blocks=[{"type": "section", "text": {"type": "mrkdwn", "text": sent_str}}],
```
JamesGlover, line 19: Please tell me that we were populating slack with "Danny Torrence left the following review for your property:"

tests/test_helpers.py
```diff
@@ -59,6 +74,23 @@ def test_add_extra_fields(centre_details):
...
+def test_check_for_required_fields(config):
+    with pytest.raises(CentreFileError, match=r".* missing in CSV file"):
+        with StringIO() as fake_csv:
+            fake_csv.write("id,RNA ID\n")
+            fake_csv.write("1,RNA_0043_\n")
+            fake_csv.seek(0)
+
+            csv_to_test_reader = DictReader(fake_csv)
+
+            assert check_for_required_fields(csv_to_test_reader, {"barcode_field": "RNA ID"}) == []
```
JamesGlover, line 96: No tests for merge_daily_files itself? Is this because its tricky to direct the tests to use a dummy downloads directory?

crawler/helpers.py
pjvv: Good spot, I included an old docstring while updating the method.

crawler/helpers.py
pjvv: Never thought of that. I think it is negligible. Each regex in the loop is potentially different so we would need to compile it.

tests/test_helpers.py
pjvv: Nope, just didn't get round to it. Will try add some.

crawler/helpers.py
pjvv: Yes, good spot thanks!

Issue #50, opened 2020-06-09
GPL-531 Add Cambridge-az plate maps to core LIMS  


rl15:
**User story**
GPL-531 | As co-coordinator for lighthouse labs (Cristina A) I would like plate maps from Cambridge-az to be parsed to support generating LH +ves on site and as sample metadata for sequencing

**Who are the primary contacts for this story**
Rich C
Siobhan AG
James G

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] .
- [ ] create yymmdd_hhmm_master.csv in psd-lims sub folder

**Additional context**
Need to establish conversation with POC cambridge, that will send incremental updates.

rl15:
Looking at barcode prefixes as part of another unrelated task noticed barcode was prefix was not specified. Thanks for not colliding with CAMB namespace in from consortium https://ssg-confluence.internal.sanger.ac.uk/x/vZYKBg . Can we change it from CB to CBAZ. Apologies missed this.

JamesGlover:
Updated the prefix, and added a bit of documentation to clarify what some of the fields do.

JamesGlover:
Seeing files ending in _Ct.csv that otherwise match the regexp. Currently we're picking these up, and then choking when we parse the date time. Need to know what the intent was for these files before fixing the issue.

JamesGlover:
Was mistake, was files with a 9999 timestamp created in the fix here (https://github.com/sanger/crawler/issues/50#issuecomment-653092341).
Should be safe to release tomorrow along with a file rename.

PR #53, opened 2020-07-02
Gpl534 only rename collections post import

### LINKED ISSUE #38: GPL-534 Rename samples collection after successful process
**User story**
Currently samples collection are recreated every day from lighthouse files. The way it works, it could happen that during an error in process because of wrong inputs, all samples from a center might not be loaded, so we could end up with a smaller set of samples in database than what we had yesterday.

Currently we keep a copy for every day of all samples loaded. It would be a more consistent approach to not change the current samples collection unless we are confident on the consistency of data loaded. We could do this by not renaming the current samples collection unless all data load have happened without any critical errors.

Eduardo Martin

**Acceptance criteria**
- [ ] Use the current version of samples collection when new version load has failed with an error

**Dependencies**

**Additional context**
See problem from load on Jun 11th

JamesGlover:
Closes #38

* Improves test coverage of main.run
* Existing sample collection only replace by new collection if import happens without critical errors


JamesGlover:
Just noticed this contains #50 (GPL-531) as well. Happy to rebase to remove it if you prefer, but its pretty much just config for the new centre.

JamesGlover:
Rebase to remove cambridge az

JamesGlover:
And of course, the rebase has broken the tests which assumed the cambridge-az files were in place.

crawler/db.py
```diff
@@ -80,24 +89,77 @@ def get_mongo_collection(database: Database, collection_name: str) -> Collection
...
-    cloned_collection = f"{collection.name}_{datetime.now().strftime('%y%m%d_%H%M')}"
+    new_name = f"{collection.name}_{suffix}"
 
-    logger.debug(f"Copying '{collection.name}' to '{cloned_collection}'")
+    rename_collection(database, collection, new_name)
+    return None
+
+
+def rename_collection(database: Database, collection: Collection, new_name: str) -> None:
+    """Renames a collection to the new name.
```
emrojo, line 108: Is database argument needed?

crawler/db.py
```diff
@@ -80,24 +89,77 @@ def get_mongo_collection(database: Database, collection_name: str) -> Collection
...
+    cleaned up to assist with debugging.
+
+    Arguments:
+        database {Database} -- the database of the collection to replace
+        collection {Collection} -- the collection to replace
+        timestamp {str} -- A timestamp to apply to the original and temporary collection names
+    """
+    temporary_collection_name = f"tmp_{collection_name}_{timestamp}"
+    logger.debug(f"Generating temporary collection: {temporary_collection_name}")
+    temporary_collection = get_mongo_collection(database, temporary_collection_name)
```
emrojo, line 143: Is get_mongo_collection creating a new collection if it doesnt exist?

crawler/main.py
```diff
@@ -62,89 +64,88 @@ def run(sftp: bool, settings_module: str = "") -> None:
...
+                logger.debug(f"Creating compund index on '{samples_collection.full_name}'")
+                # create compound index on 'Root Sample ID', 'RNA ID', 'Result', 'Lab ID' - some data
+                #   had the same plate tested at another time so ignore the data if it is exactly the
+                #   same
+                samples_collection.create_index(
+                    [
+                        (FIELD_ROOT_SAMPLE_ID, pymongo.ASCENDING),
+                        (FIELD_RNA_ID, pymongo.ASCENDING),
+                        (FIELD_RESULT, pymongo.ASCENDING),
+                        (FIELD_RNA_ID, pymongo.ASCENDING),
```
harrietc52, line 82: `FIELD_RNA_ID` is duplicate with line 80

crawler/db.py
JamesGlover: Good catch. Nope, doesn't appear to be, think I was using the method signature of the original copy approach.

crawler/db.py
JamesGlover: Sort of, Mongo creates collections lazily as soon as something gets added to them. Its actually a bit frustrating, as it makes it surprisingly tricky to check if a collection already exists or not.

crawler/main.py
JamesGlover: Thanks, order of indicies is important, but it looks like the first definition is taking precedence here (based on observing the actual index that gets created) so I should be able to drop the second index.

PR #86, opened 2020-07-30
Reenable cambridge az

emrojo:
Reenable cambridge az
Added current blacklist version for Alderley

Closes GPL590

### DIFF
--- a/crawler/config/defaults.py
+++ b/crawler/config/defaults.py
@@ -44,6 +44,8 @@ CENTRES = [
             "AP_sanger_report_200725_0001.csv",
             "AP_sanger_report_200726_0001.csv",
             "AP_sanger_report_200727_0001.csv",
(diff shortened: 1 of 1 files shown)


crawler/config/defaults.py
```diff
@@ -45,6 +45,7 @@
             "AP_sanger_report_200726_0001.csv",
             "AP_sanger_report_200727_0001.csv",
             "AP_sanger_report_200728_0001.csv",
+            "AP_sanger_report_200729_0001.csv",
         ],
```
JamesGlover, line 49: Is this intentional?

JamesGlover: Via slack: yes, that was the change from yesterday
14:48
we changed it inside the container, but with this we update in the code, so if someone deploys we dont lose the change

PR #101, opened 2020-08-11
Gpl577 refactor to persist database

emrojo:
Closes GPL577

* Process files and keep databse


emrojo:
Checklist to create volume:
[ ] In psd-provision select the workspace (training)
[ ] terraform apply (to create the volume; we dont have to restart the server, is automatically added as a new device (/dev folder)), to know the name check in ```dmesg``` output
[ ] Log into the machine and create the filesystem in the new added device
   ```sudo mkfs.ext4 /dev/disk/by-id/virtio-cbec7cb1-***-***-***```
[ ] Create a new folder to mount the filesystem
  ```sudo mkdir /var/lib/crawler```
[ ] Edit /etc/fstab and add the mount of the new filesystem in the new created folder
(It should do something like this):
  ```sudo mount /dev/disk/by-id/virtio-cbec7cb1-***-***-*** /var/lib/crawler```
line like: 
```/dev/vdb	/var/lib/crawler	ext4	defaults	0 0```

PR #109, opened 2020-09-09
Gpl579 write to mlwh

KatyTaylor:
ready for review


run_migration.py
```diff
@@ -1,4 +1,46 @@
...
+    print("Running update_mlwh_with_legacy_samples migration")
+    update_mlwh_with_legacy_samples.run(s_start_datetime=s_start_datetime, s_end_datetime=s_end_datetime)
+
+def migration_by_name(migration_name):
+    switcher = {
+        'sample_timestamps': migration_sample_timestamps,
+        'update_mlwh_with_legacy_samples': migration_update_mlwh_with_legacy_samples,
+    }
+    # Get the function from switcher dictionary
+    func = switcher.get(migration_name, lambda: print("Invalid migration name, aborting"))
```
KatyTaylor, line 37: So does it actually abort the program here, if the key doesn't exist? If so, what makes it do that rather than carry on to `func()` ?

tests/migrations/test_update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,162 @@
...
+import pytest
+from datetime import (datetime, timedelta)
+
+def print_exception(e: Exception) -> None:
+    print(f'An exception occurred, at {datetime.now()}')
+    e = sys.exc_info()
+    print(e[0]) # exception type
+    print(e[1]) # exception message
+    if e[2]: # traceback
+      traceback.print_tb(e[2], limit=10)
```
KatyTaylor, line 29: So I think we could just import logger and do `logger.exception(e)` - did you try it?

tests/migrations/test_update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,162 @@
...
+    FIELD_DATE_TESTED,
+    FIELD_LAB_ID,
+    FIELD_SOURCE,
+    FIELD_PLATE_BARCODE,
+    FIELD_CREATED_AT,
+    FIELD_UPDATED_AT,
+    MLWH_TABLE_NAME,
+    MONGO_DATETIME_FORMAT,
+)
+import pytest
```
KatyTaylor, line 20: you don't need the above line

tests/migrations/test_update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,162 @@
...
+
+    print(samples)
+    return samples
+
+def clear_mlwh_table(mysql_conn):
+    print("Clearing MLWH database table")
+
+    cursor = mysql_conn.cursor()
+    try:
+        cursor.execute(f"TRUNCATE TABLE unified_warehouse_test.{MLWH_TABLE_NAME}")
```
KatyTaylor, line 54: Should use MLWH_DB_DBNAME, since it's in the constants?

run_migration.py
```diff
@@ -1,4 +1,46 @@
...
+print("* sample_timestamps")
+print("* update_mlwh_with_legacy_samples")
+
+def migration_sample_timestamps():
+    print("Running sample_timestamps migration")
+    sample_timestamps.run()
+
+def migration_update_mlwh_with_legacy_samples():
+    if not len(sys.argv) == 4:
+        print("Please add both start and end datetime arguments for this migration (format yymmdd_hhmm e.g. 200115_1200), aborting")
```
KatyTaylor, line 23: I like the comments. Can we add info on whether or not the start and end dates are inclusive or exclusive?

tests/migrations/test_update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,162 @@
...
+def test_valid_datetime_string_valid(config):
+    valid_dt = datetime.strftime(datetime.now(), MONGO_DATETIME_FORMAT)
+    assert(valid_datetime_string(valid_dt) == True)
+
+def test_basic_usage(mongo_database, mlwh_connection):
+    config, mongo_db = mongo_database
+
+    start_datetime = datetime(2020, 5, 10, 15, 10)
+
+    # generate and insert sample rows into the mobgo database
```
KatyTaylor, line 78:
```suggestion
    # generate and insert sample rows into the mongo database
```
```suggestion
    # generate and insert sample rows into the mobgo database
```

tests/migrations/test_update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,162 @@
...
+    clear_mlwh_table(mysql_conn)
+
+    s_start_datetime = datetime.strftime(start_datetime, MONGO_DATETIME_FORMAT)
+    s_end_datetime = datetime.strftime(start_datetime + timedelta(days=3), MONGO_DATETIME_FORMAT)
+
+    # run the method to update the MLWH from the mongo database
+    update_mlwh_with_legacy_samples(config, s_start_datetime, s_end_datetime)
+
+    # query for selecting rows from MLWH (it was emptied before so select * is fine for this)
+    sql_query = f"SELECT * FROM unified_warehouse_test.{MLWH_TABLE_NAME} ORDER BY {FIELD_CREATED_AT} ASC"
```
KatyTaylor, line 98: Same as above - use MLWH_DB_DBNAME?

KatyTaylor: Should be MLWH_CREATED_AT ?

KatyTaylor: Should be MLWH_CREATED_AT ?

tests/migrations/test_update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,162 @@
...
+
+    try:
+        # run the query and fetch the results
+        cursor = mysql_conn.cursor()
+        cursor.execute(sql_query)
+        records = cursor.fetchall()
+
+        # check there are the expected number of rows in the MLWH (4?)
```
KatyTaylor, line 106:
```suggestion
        # check there are the expected number of rows in the MLWH (only 4 of 6 fell within date range)
```

tests/migrations/test_update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,162 @@
...
```
KatyTaylor, line 154:
```suggestion
        # check there are the expected number of rows in the MLWH (0 fell within date range)
```

tests/migrations/test_update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # use timestamps such that no rows qualify
+    s_start_datetime = datetime.strftime(start_datetime + timedelta(days=6), MONGO_DATETIME_FORMAT)
+    s_end_datetime = datetime.strftime(start_datetime + timedelta(days=8), MONGO_DATETIME_FORMAT)
+
+    # run the method to update the MLWH from the mongo database
+    update_mlwh_with_legacy_samples(config, s_start_datetime, s_end_datetime)
+
+    # query for selecting rows from MLWH (it was emptied before so select * is fine for this)
+    sql_query = f"SELECT * FROM unified_warehouse_test.{MLWH_TABLE_NAME} ORDER BY {FIELD_CREATED_AT} ASC"
```
KatyTaylor, line 146: Same comments as same line in previous test

migrations/update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,20 @@
+from crawler.helpers import (
+    get_config
+)
+from datetime import datetime
+from migrations.helpers import mlwh_samples_update_helper
+# import sys
```
KatyTaylor, line 6:
```suggestion
```

migrations/update_mlwh_with_legacy_samples.py
```diff
@@ -0,0 +1,20 @@
...
+
+# Start and End datetime parameters must be in format YYYY MM DD HH mm
```
KatyTaylor, line 8:
```suggestion
# Start and End datetime parameters must be in format YYYYMM_DDHHmm
```
Is this correct?

migrations/helpers/mlwh_samples_update_helper.py
```diff
@@ -0,0 +1,104 @@
...
+)
+from crawler.db import (
+    create_mongo_client,
+    get_mongo_collection,
+    get_mongo_db,
+    create_mysql_connection,
+    run_mysql_executemany_query,
+)
+from crawler.helpers import (
+    get_config,
```
KatyTaylor, line 21: not used

migrations/helpers/mlwh_samples_update_helper.py
```diff
@@ -0,0 +1,104 @@
...
+        print_exception()
+        return False
+
+def print_exception() -> None:
+    print(f'An exception occurred, at {datetime.now()}')
+    e = sys.exc_info()
+    print(e[0]) # exception type
+    print(e[1]) # exception message
+    if e[2]: # traceback
+      traceback.print_tb(e[2], limit=10)
```
KatyTaylor, line 43: like I said somewhere else, could replace with logger.exception(e)

migrations/helpers/mlwh_samples_update_helper.py
```diff
@@ -0,0 +1,104 @@
...
+
+def update_mlwh_with_legacy_samples(config, s_start_datetime: str = "", s_end_datetime: str = "") -> None:
+    if not valid_datetime_string(s_start_datetime):
+        print('Aborting run: Start datetime is not in the expected datetime format')
```
KatyTaylor, line 47: Format should be ...

migrations/helpers/mlwh_samples_update_helper.py
```diff
@@ -0,0 +1,104 @@
...
+    start_datetime = datetime.strptime(s_start_datetime, MONGO_DATETIME_FORMAT)
+    end_datetime = datetime.strptime(s_end_datetime, MONGO_DATETIME_FORMAT)
+
+    if start_datetime > end_datetime:
+        print("Aborting run: End datetime must be greater than start datetime")
+        return
+
+    print(f"Starting MLWH update process with start datetime {start_datetime} and end datetime {end_datetime}")
+
+    try:
```
KatyTaylor, line 63: hmm, not sure about having the try / except here, what does it give us?

emrojo (changes requested):

crawler/sql_queries.py
```diff
@@ -0,0 +1,67 @@
...
+`created_at_external` datetime DEFAULT NULL COMMENT 'When the corresponding record was inserted into the MongoDB',
+`updated_at_external` datetime DEFAULT NULL COMMENT 'When the corresponding record was last updated in MongoDB',
+`created_at` datetime DEFAULT NULL COMMENT 'When this record was inserted',
+`updated_at` datetime DEFAULT NULL COMMENT 'When this record was last updated',
+PRIMARY KEY (`id`),
+UNIQUE KEY `index_lighthouse_sample_on_root_sample_id_and_rna_id_and_result` (`root_sample_id`,`rna_id`,`result`),
+UNIQUE KEY `index_lighthouse_sample_on_mongodb_id` (`mongodb_id`),
+KEY `index_lighthouse_sample_on_date_tested` (`date_tested`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
+"""
```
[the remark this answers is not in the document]
emrojo: This is more for future improvements. Not for current version.

emrojo (changes requested):

crawler/file_processing.py
```diff
@@ -439,9 +464,49 @@ def insert_samples_from_docs(self, docs_to_insert) -> None:
...
+
+        if mysql_conn is not None and mysql_conn.is_connected():
+            try:
+                run_mysql_executemany_query(mysql_conn, SQL_MLWH_MULTIPLE_INSERT, values)
+
+                logger.debug(f"MLWH database inserts completed successfully for file {self.file_name}")
+            except Exception as e:
+                self.logging_collection.add_error(
+                    "TYPE 14",
+                    f"MLWH database inserts failed for file {self.file_name}",
```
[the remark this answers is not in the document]
emrojo: As we have a procedure, maybe add it to the README about how we can update mlwh using the range of dates.

emrojo (changes requested):

crawler/file_processing.py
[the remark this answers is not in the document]
andrewsparkes: Added to ReadMe

migrations/helpers/mlwh_samples_update_helper.py
andrewsparkes: Removed

migrations/helpers/mlwh_samples_update_helper.py
andrewsparkes: I was using print here because I copied the samples migration. This migration is run at the command line, so maybe logger.exception is not ideal as not visible?

KatyTaylor: I wasn't sure if logger messages would be visible or not - have you tested it? Of course, if they are not visible then stick with print!

andrewsparkes: Tried it and they don't show.

migrations/helpers/mlwh_samples_update_helper.py
andrewsparkes: Changed

migrations/helpers/mlwh_samples_update_helper.py
KatyTaylor: :(

migrations/helpers/mlwh_samples_update_helper.py
andrewsparkes: Catches and prints nicely any exception thrown from the database connections.

migrations/update_mlwh_with_legacy_samples.py
andrewsparkes: Removed

migrations/update_mlwh_with_legacy_samples.py
andrewsparkes: Changed

migrations/helpers/mlwh_samples_update_helper.py
KatyTaylor: Hmm ok, if the exception was thrown it would print it out anyway - but maybe it's easier to read this way?

run_migration.py
andrewsparkes: Added. And there's a new section in the Readme about how to run it too.

run_migration.py
andrewsparkes: I think the way it works is it either sets the func to the method named in the switcher, or if that doesn't return anything it sets it to the lambda print('stuff') function. So func() runs either way, it just prints that message if the migration name is not valid.

tests/migrations/test_update_mlwh_with_legacy_samples.py
andrewsparkes: import pytest? are you sure, it's in all the other test files

KatyTaylor: Dunno, I took it out and ran it and it seemed to work. It's in conftest.py so maybe that's why?

tests/migrations/test_update_mlwh_with_legacy_samples.py
andrewsparkes: using pytest.fail() instead

tests/migrations/test_update_mlwh_with_legacy_samples.py
andrewsparkes: Truncate table wipes the table of all data. I thought that would be very bad if it was run on prod, so hardcoded the test database.. What do you think though? any way a test can be run against anything other than a test database?

tests/migrations/test_update_mlwh_with_legacy_samples.py
andrewsparkes: Removed, seems to run still

tests/migrations/test_update_mlwh_with_legacy_samples.py
andrewsparkes: Changed

tests/migrations/test_update_mlwh_with_legacy_samples.py
andrewsparkes: Changed

tests/migrations/test_update_mlwh_with_legacy_samples.py
andrewsparkes: Changed

README.md
```diff
@@ -58,8 +58,30 @@ The following runtime flags are available:
...
+When the crawler process runs nightly it should be updating the MLWH lighthouse_sample table as it goes with records for all rows that are inserted into MongoDB.
+If that MLWH insert process fails you should see a critical exception for the file in Lighthouse-UI. This may be after records inserted correctly into MongoDB, and re-running the file will not re-attempt the MLWH inserts in that situation.
+There is a manual migration task that can be run to fix this discrepancy (update_mlwh_with_legacy_samples) that allows insertion of rows to the MLWH between two MongoDB created_at datetimes.
+NB. Both datetimes are inclusive: range includes those rows greater than or equal to start datetime, and less than or equal to end datetime.
+
+Usage (inside pipenv shell):
+
+    $ python run_migration.py update_mlwh_with_legacy_samples 200115_1200 200116_1600
+
+Where the time format is YYMMDD_HHmm. Both start and end timestamps nust be present.
```
KatyTaylor, line 72:
```suggestion
Where the time format is YYMMDD_HHmm. Both start and end timestamps must be present.
```

README.md
```diff
@@ -58,8 +58,30 @@ The following runtime flags are available:
...
+
+    $ python run_migration.py update_mlwh_with_legacy_samples 200115_1200 200116_1600
+
+Where the time format is YYMMDD_HHmm. Both start and end timestamps must be present.
+
+The process should not duplicate rows that are already present in MLWH, so you can be generous with your timestamp range.
+
 ## Testing
 
+The tests require a connection to the 'lighthouse_samples' table in the Multi-LIMS Warehouse.
```
KatyTaylor, line 78:
```suggestion
The tests require a connection to the 'lighthouse_sample' table in the Multi-LIMS Warehouse.
```

crawler/file_processing.py
```diff
@@ -439,9 +464,49 @@ def insert_samples_from_docs(self, docs_to_insert) -> None:
 
             self.docs_inserted = e.details["nInserted"]
             self.add_duplication_errors(e)
+
+            errored_ids = list(map(lambda x: x['op'][FIELD_MONGODB_ID], e.details["writeErrors"]))
+            inserted_ids = [ doc[FIELD_MONGODB_ID] for doc in docs_to_insert if doc[FIELD_MONGODB_ID] not in errored_ids ]
```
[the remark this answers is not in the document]
KatyTaylor: Hi @emrojo , we haven't addressed this comment. I'm not sure what other errors we'd expect. This is in a block where we've caught a 'BulkWriteError', so hopefully those errors should be in the 'writeErrors' section?

If there is a different kind of exception thrown, it will get caught in the below block, so we should get to know about it.

You are right, it's possible this would end up causing a discrepancy between the 2 dbs. I'd rather not have to requery the ids from MongoDB though, as I think it would be a v large query statement, which Mongo doesn't like, and would slow down the process.

I'm tempted to leave it, and amend it in future if we become aware of other forms of errors. Do you think it's OK to merge in this state?

andrewsparkes (approved):

crawler/file_processing.py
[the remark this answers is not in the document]
emrojo: Yes, I agree with that, is also a type of error we cannot predict, ok to merge

emrojo (approved):

PR #120, opened 2020-09-25
Gpl579 write to mlwh - date fixes

KatyTaylor:


crawler/helpers.py
```diff
@@ -160,7 +156,7 @@ def map_lh_doc_to_sql_columns(doc) -> Dict[str, Any]:
             Dict[str, str] -- Dictionary of MySQL versions of fields
     """
     value = map_mongo_to_sql_common(doc)
-    dt = datetime.strftime(datetime.now(timezone.utc), MYSQL_DATETIME_FORMAT)
+    dt = datetime.now(timezone.utc)
```
andrewsparkes, line 159: ah, so you don't need to format it for mysql at all, it just takes the datetime as is? neat

crawler/helpers.py
```diff
@@ -175,11 +171,11 @@ def map_mongo_doc_to_sql_columns(doc) -> Dict[str, Any]:
             Dict[str, str] -- Dictionary of MySQL versions of fields
     """
     value = map_mongo_to_sql_common(doc)
-    value[MLWH_CREATED_AT] = datetime.strftime(doc[FIELD_CREATED_AT], MYSQL_DATETIME_FORMAT)
-    value[MLWH_UPDATED_AT] = datetime.strftime(doc[FIELD_UPDATED_AT], MYSQL_DATETIME_FORMAT)
+    value[MLWH_CREATED_AT] = doc[FIELD_CREATED_AT]
```
andrewsparkes, line 174: So here the doc has a datetime value, so you can just pass that to the MLWH as is? great

crawler/helpers.py
```diff
@@ -190,16 +186,9 @@ def parse_date_tested(date_string: str) -> str:
...
-        return ''
-
-    if date_string.find('UTC') != -1:
-        # timezone doesn't get set despite the '%Z' in the format string, unless we do this
-        date_time = date_time.replace(tzinfo=timezone.utc)
-
-    date_time_str = datetime.strftime(date_time, MYSQL_DATETIME_FORMAT)
-
-    return date_time_str
+        return None
```
andrewsparkes, line 191: mypy issue, return type is datetime

tests/test_file_processing.py
```diff
@@ -614,7 +630,7 @@ def test_set_state_for_file_when_in_success_folder(config):
     return False
 
 # tests for inserting docs into mlwh
-def test_insert_samples_from_docs_into_mlwh(config):
+def test_insert_samples_from_docs_into_mlwh(config, mlwh_connection):
```
andrewsparkes, line 633: Ok so now you're clearing the table at the start of each test by doing it in the mlwh_connection method? then leaving that connection open while testing. If the test fails the finally still closes the mlwh_connection? and the tested code is opening and handling it's own separate connection. Seems neat.

andrewsparkes (commented):
mypy issue needs fix else good to go

andrewsparkes (approved):

PR #124, opened 2020-10-02
GPL642 and 654 adding crawler lighthouse CT columns

### LINKED ISSUE #108: GPL-642 Parsing and persisting CT values from AP
**User story**
GPL-642 | As senior stakeholders (Jeff B & John S) we would like to persist CT values for Alderney Park to later filter out these samples from sentinel to increase sequencing success rates

Sonia G
Cristina A
Jeff B
John S (PAM)
Rob A
Emma G

**Candidate Acceptance criteria**
- [ ] allow current file standard and new file standard to be both valid (LH labs will all migrate to this new standard but not as one entity/ on same day
- [ ] parse the file, reject & log any invalid entries using following criteria;

Jeff B wrote (Wednesday, 9 September 2020 at 10:59)

To answer your questions:
 
For the Cq columns, the valid range is a float between 0 and 100, can be null 
For the Result columns, the valid values are Positive and Negative, can be null
For the Target columns, this is a string with currently one of only four values  {‘ORF1ab’, ‘N gene’, ‘S gene’, ‘MS2’} . 
It may be sensible to make these the only valid values, and then if the labs change Targets without telling us it will trigger an exception. Though it will be important to be able to add new Targets in the future.
 
- [ ] Log invalid an sample when
 
Result (in col E) = ‘Positive’ but **all** CHn-Result = ‘Negative’
CT value  less than 0 or greater than 100
CHn_Target value not in set {‘ORF1ab’, ‘N gene’, ‘S gene’, ‘MS2’} ? NB This set will have further elements in future (suggest is lightweight to add new elements

- [ ] persist file in mongo database
- [ ] persist new data in MLWH (see GPL-579)

**File standard changes**

Current AP file format
(cols A-G)
Root Sample ID	
Viral Prep ID	
RNA ID	
RNA-PCR ID	
Result	
Date Tested	
Lab ID

extended by new file headers

(cols H-S)
CH1-Target  {‘ORF1ab’}
CH1-Result {‘Positive’, ‘Negative’, _null_} 
CH1-Cq {(0..40), _null_}
CH2-Target {‘N gene’}
CH2-Result{‘Positive’, ‘Negative’, _null_}
CH2-Cq {(0..40), _null_}
CH3-Target {‘S gene’}
CH3-Result {‘Positive’, ‘Negative’, _null_}
CH3-Cq {(0..40), _null_}
CH4-Target {‘MS2’}
CH4-Result {‘Positive’, ‘Negative’, _null_}
CH4-Cq {(0..40), _null_}

See example file below (zipped)

**Additional context**
Business impact will be in at least three parts

1. Parsing the files for AP and persisting the CT values (this backlog item)
2. Repeating for one or more of the other LH sites (Agreed to do AP first in project meeting on 23rd Sept)
3. Using this data to filter out any results = positive from '+ves on site LH report' available here: http://lighthouse-ui.psd.sanger.ac.uk/ (maybe this backlog item or another -to discuss esp reporting) 

**Team discussion**

Discussed post stand up on the 24th with reference to this diagram: https://app.lucidchart.com/invitations/accept/6499b3a0-7a14-4192-b5db-1cd35b84646f 

Agreed best approach is to, parse and persist first (in MLWH) then filter as a separate backlog item (GPL-659)
 
Even if we are treating these like negatives for Cherry picking, they should not be seen as negatives from a reporting context

We would expect that the API call from Beckman robotics of +ve wells (GPL-567 & 568), that high CT values will be filtered out from automated picks.

### LINKED ISSUE #115: GPL-654 Parse & persist new result value 'limit of detection'
**User story**
GPL-654 | As senior stakeholders (John S & Sonia G) we would like new covid results value to be parsed and persisted rather than rejected as invalid

John S 
Sonia G

**Question**
Q. What would happen now, sample ignored and so would not be picked? - crucially does it fail safely today?
A. From discussion on 24th Sept will fails safe however sample will not be persisted in MLWH without this backlog item  

**Acceptance criteria**
- [ ] Parse new results value 'limit of detection'
- [ ] Persist data in crawler db
- [ ] Persist data to MLWH (GPL-579)
- [ ] update documentation if required: https://ssg-confluence.internal.sanger.ac.uk/x/jqsKBg 

**Dependencies**
This story may have dependency or advantage in doing alongside GPL-642

**Additional context**
From go live to date covid result has been one of four values {
Positive,
Negative,
Void
_null_
}

LH labs expecting to introduce new value 'limit of detection'

Discussed post stand up on the 24th with reference to this diagram: https://app.lucidchart.com/invitations/accept/6499b3a0-7a14-4192-b5db-1cd35b84646f

andrewsparkes:
Closes #108 
Closes #115


crawler/constants.py
```diff
@@ -37,9 +49,36 @@
...
+MLWH_CH4_CQ = "ch4_cq"
 MLWH_CREATED_AT = "created_at"
 MLWH_UPDATED_AT = "updated_at"
 
 # datetime formats
 MONGO_DATETIME_FORMAT = "%y%m%d_%H%M"
 MYSQL_DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S"
+
+# allowed Result field values
+# TODO: confirm! 'Detected', 'Not detected'?
```
KatyTaylor, line 72: Still got a TODO here

andrewsparkes: Yeah it's still TODO, Rich is clarifying

crawler/constants.py
```diff
@@ -37,9 +49,36 @@
...
+
+# allowed CT channel CHn-Target field values (or can be null)
+ALLOWED_CH_TARGET_VALUES = ('ORF1ab', 'N gene', 'S gene', 'MS2')
+
+# allowed CT channel CHn-Result field values (or can be null)
+ALLOWED_CH_RESULT_VALUES = ('Positive', 'Negative', 'Inconclusive', 'Void')
+
+# range of allowed cq values (0 .. 100, set as strings for conversion to decimals in code)
+MIN_CQ_VALUE = '0.0'
+MAX_CQ_VALUE = '100.0'
```
KatyTaylor, line 83: Small thing, but I think I would personally store them as decimals here:
e.g. `MIN_CQ_VALUE = Decimal('0.0')`
and then make the `is_within_cq_range` method expect 3 decimals

crawler/file_processing.py
```diff
@@ -536,6 +572,14 @@ def get_required_headers(self) -> Set[str]:
 
         return required
 
+    def get_channel_headers(self) -> Set[str]:
+        """Rreturns the list of optional headers.
```
KatyTaylor, line 576:
```suggestion
        """Returns the list of optional headers.
```

crawler/file_processing.py
```diff
@@ -536,6 +572,14 @@ def get_required_headers(self) -> Set[str]:
...
+
+            Returns:
+                [str] - array of headers
```
KatyTaylor, line 579: But it returns a set, not an array

crawler/file_processing.py
```diff
@@ -639,10 +683,16 @@ def filtered_row(self, row, line_number) -> Dict[str, str]:
                 self.log_adding_default_lab_id(row, line_number)
 
         # filter out any unexpected columns
+        # get values for required headers
```
KatyTaylor, line 686:
```suggestion
        # get names of required column headers
```

andrewsparkes: We're getting the values from those column fields in the file row, and adding them into the modified row dict.
I guess it's k,v pairs really

crawler/file_processing.py
andrewsparkes: Fixed

crawler/file_processing.py
andrewsparkes: Fixed (and in required headers one)

crawler/constants.py
andrewsparkes: Changed.

crawler/file_processing.py
KatyTaylor: oh I see, I misread it as just referring to the get_required_headers method, but it's referring to what the whole block does

andrewsparkes: I'll make it clearer then

andrewsparkes: Done

crawler/file_processing.py
```diff
@@ -663,60 +717,394 @@ def format_and_filter_rows(self, csvreader: DictReader) -> Any:
...
+        # add file details and import timestamp
+        import_timestamp = self.get_now_timestamp()
+
+        modified_row[FIELD_LINE_NUMBER] = line_number  # type: ignore
+        modified_row[FIELD_FILE_NAME] = self.file_name
+        modified_row[FIELD_FILE_NAME_DATE] = self.file_name_date()
+        modified_row[FIELD_CREATED_AT] = import_timestamp
+        modified_row[FIELD_UPDATED_AT] = import_timestamp
+
+        # store row signature to allow checking for duplicates in follwing rows
```
KatyTaylor, line 821:
```suggestion
        # store row signature to allow checking for duplicates in following rows
```

crawler/file_processing.py
```diff
@@ -663,60 +717,394 @@ def format_and_filter_rows(self, csvreader: DictReader) -> Any:
...
-                        self.logging_collection.add_error(
-                            "TYPE 5",
-                            f"Duplicated, line: {line_number}, root_sample_id: {row[FIELD_ROOT_SAMPLE_ID]}",
-                        )
-                        continue
-                    seen_rows.add(row_signature)
-                    augmented_data.append(row)
-                else:
-                    missing_data_count += 1
+        # check if the ct channel result values match to expected (if present)
```
KatyTaylor, line 788: remove comment above and equivalent ones?  I think the method names are v self-explanatory

crawler/file_processing.py
```diff
@@ -663,60 +717,394 @@ def format_and_filter_rows(self, csvreader: DictReader) -> Any:
...
+        modified_row[FIELD_FILE_NAME_DATE] = self.file_name_date()
+        modified_row[FIELD_CREATED_AT] = import_timestamp
+        modified_row[FIELD_UPDATED_AT] = import_timestamp
+
+        # store row signature to allow checking for duplicates in follwing rows
+        seen_rows.add(row_signature)
+
+        return modified_row
+
+    def row_result_value_valid(self, row, line_number) -> bool:
```
KatyTaylor, line 826: Agreed, we could compact down this validation. 
Might be nice to move it into a separate field_validation.py file too, or similar? See what you think after you've done your DRYing of code.

crawler/file_processing.py
andrewsparkes: Fixed

crawler/file_processing.py
andrewsparkes: Removed

crawler/file_processing.py
```diff
@@ -663,60 +717,394 @@ def format_and_filter_rows(self, csvreader: DictReader) -> Any:
...
+                self.logging_collection.add_error(
+                    "TYPE 18",
+                    f"CH4-Result invalid, line: {line_number}, result: {row[FIELD_CH4_RESULT]}",
+                )
+                return False
+
+        return True
+
+
+    def is_number(self, item):
```
KatyTaylor, line 934: Think you could slim this down by 
a) Get rid of isinstance() check - assume the input is a string, let it blow up if it's not (it will always be in our case right)
b) `return (decimal_pattern.match(item) or int_pattern.match(item))` instead of the if / else / return ?

Another option is to try / except casting to a Decimal

crawler/file_processing.py
```diff
@@ -663,60 +718,395 @@ def format_and_filter_rows(self, csvreader: DictReader) -> Any:
...
+        Arguments:
+            row {Row} - row object from the csvreader
+            line_number {integer} - line number within the file
+
+        Returns:
+            bool - whether the channel results complement the main results
+        """
+        # if the result is not positive we do not need to check any further
+        if row[FIELD_RESULT] != 'Positive':
+            return True
```
KatyTaylor, line 1071: OK still got a few comments about this validation, sorry! (just from reading JB's comments on the story)

1.
Think another disallowed combo would be if: \
Result = 'Negative'
Any of the CHn-Result values = 'Positive'
?

2. Don't think you should be using the FIELD_CH4_RESULT at all, just 1-3

KatyTaylor (dismissed):

PR #148, opened 2020-11-10
Gpl708 dart inserts

### LINKED ISSUE #145: GPL-708 As developers we want to ensure that when a plate map is processed data is inserted into the DART database
**User story**
As developers we want to ensure that when a plate map is processed data is inserted into the DART database including all the plate barcodes (whether contain +ves or not), and any wells that we want to pick (based on existing query)

**Acceptance criteria**
Data inserted into the DART database includes:
- [x] All 'new' plates, with barcodes
- [x] Wells on those plates that we want to pick, based on current criteria (see `POSITIVE_SAMPLES_MONGODB_FILTER` in `constants.py` in lighthouse repo) (see https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/UAT+6th+October+2020 for further examples and info about the filtering rules)
- [x] (*if Option 1 is chosen [here](https://app.lucidchart.com/invitations/accept/869a412c-ed4a-406a-a7b8-ec7bbd4f97b1)*) Sample metadata on the above wells, including: 
    - Root Sample ID ✅
    - Result _Only RESULT=positive_ samples are added, with a **state** that shows whether the plate is 'pickable' or 'picked' ✅
    - ~~COG UK ID (needs a decision - depends when we generate it)~~ Not needed at this stage in the process. See below
    - RNA ID (needs a decision - we could reconstruct this later based on plate barcode and well position) ✅
    - LAB ID ✅
- [] ~~(*up for discussion*) - also create a COG UK ID at this point using _baracoda_ and update MLWH with it~~ Will instead be done as part of [GPL-710](https://github.com/sanger/lighthouse/issues/145)
- [x] set the _biomek_labclass_ parameter to a value that we will store in the centre configuration in `defaults.py` 

**Additional context**
This will use the DART stored procedures to insert / update plates and well information.

Needs to be able to cope with re-uploads of files where they may fix errored rows, potentially adding more positives to pick, whole point of error reporting is to give them chance to do that.

See diagram here - https://app.lucidchart.com/invitations/accept/5285ec96-5090-4245-8d9b-bc9d671bb49a

Chris-Friend:

* Allow creating a connection to the DART database
* Add new errors (TYPEs 22, 23, 24)
* Add RESULT=positive samples that are not in Mongo to DART
* In adding to DART:
  * Do not override existing plates
  * Do not allow updating plates not in the 'pending' state
  * Handle unexpected inconsistencies in DART (a plate exists but does not have a state)

closes #145


crawler/db.py
```diff
@@ -259,3 +268,72 @@ def init_warehouse_db_command():
...
+    if readonly:
+        dart_db_username = config.DART_DB_RO_USER  # type: ignore
+        dart_db_password = config.DART_DB_RO_PASSWORD  # type: ignore
+    else:
+        dart_db_username = config.DART_DB_RW_USER  # type: ignore
+        dart_db_password = config.DART_DB_RW_PASSWORD  # type: ignore
+    dart_db_db = config.DART_DB_DBNAME  # type: ignore
+    dart_db_driver = config.DART_DB_DRIVER  # type: ignore
+
+    connection_string = 'DRIVER='+dart_db_driver+';SERVER='+dart_db_host+f';PORT={dart_db_port};DATABASE='+dart_db_db+';UID='+dart_db_username+';PWD=' + dart_db_password
```
JamesGlover, line 293:
```suggestion
    connection_string = f'DRIVER={dart_db_driver};SERVER={dart_db_host};PORT={dart_db_port};DATABASE={dart_db_db};UID={dart_db_username};PWD={dart_db_password}'
```

crawler/db.py
```diff
@@ -259,3 +268,72 @@ def init_warehouse_db_command():
...
+
+        if sql_server_conn is not None:
+            logger.debug('DART Connection Successful')
+        else:
+            logger.error('DART Connection Failed')
+
+    except Error as e:
+        logger.error(f"Exception on connecting to DART database: {e}")
+
+    return sql_server_conn
```
JamesGlover, line 309: The return type at the top of this function implies we can expect a pyodbc.Connection however it looks like we can also return None. If:
- The code throws an execption
- pyodbc.connect(connection_string) returns None

This means the should be specifying Optional[pyodbc.Connection]

Also, as best I can tell, on failed connections, pyodbc.connect(connection_string) throws, and should never return None. So the check in the middle can be removed. (Unless I'm missing something, the documentation wasn't great)

I'd probably actually throw an exception in these circumstances instead, but that would also requiring restructuring of the code elsewhere, so probably not something to do now.

crawler/file_processing.py
```diff
@@ -1043,22 +1157,22 @@ def row_positive_result_matches_channel_results(self, row, line_number) -> bool:
...
+            if row.get(FIELD_CH3_RESULT) == POSITIVE_RESULT_VALUE:
                 ch_results_positive += 1
 
         if row.get(FIELD_CH4_RESULT):
             ch_results_present += 1
-            if row.get(FIELD_CH4_RESULT) == 'Positive':
+            if row.get(FIELD_CH4_RESULT) == POSITIVE_RESULT_VALUE:
                 ch_results_positive += 1
 
         # if there are no channel results present in the row we do not need to check further
```
JamesGlover, line 1178: Definitely room for some DRYing out here, but I don't want to make you clean up every line that you touched refactoring something else.

crawler/helpers.py
```diff
@@ -485,6 +494,30 @@ def __init__(self):
         self.max_errors = 5
         self.short_display_description = "Result not aligned with CHn-Results"
 
+class AggregateType22(AggregateTypeBase):
```
JamesGlover, line 497: This might need a comment to explain what it is.

JamesGlover: Oh, okay,  github context wasn't making it clear that there are tonnes of these.

tests/test_db.py
```diff
@@ -110,3 +114,47 @@ def test_run_mysql_executemany_query_execute_error(config):
...
+
+def test_create_dart_sql_server_conn_readwrite(config):
+    with patch('pyodbc.connect') as mock_connect:
+        conn_string = f'DRIVER={config.DART_DB_DRIVER};SERVER={config.DART_DB_HOST};PORT={config.DART_DB_PORT};DATABASE={config.DART_DB_DBNAME};UID={config.DART_DB_RW_USER};PWD={config.DART_DB_RW_PASSWORD}'
+        create_dart_sql_server_conn(config, readonly=False)
+        mock_connect.assert_called_with(conn_string)
+
+def test_create_dart_sql_server_conn_none(config):
+    with patch('pyodbc.connect', return_value = None):
+        assert create_dart_sql_server_conn(config) == None
```
JamesGlover, line 132: Guess this relates to my earlier comment, is there reason to believe this happens?

tests/test_db.py
```diff
@@ -110,3 +114,47 @@ def test_run_mysql_executemany_query_execute_error(config):
...
+
+def test_create_dart_sql_server_conn_expection(config):
+    with patch('pyodbc.connect', side_effect = Exception('Boom!')):
+        with pytest.raises(Exception):
+            create_dart_sql_server_conn(config)
+
```
JamesGlover, line 138: Huh... okay, looks like I might have misunderstood python exception handling if this test is passing.

crawler/db.py
Chris-Friend: I think this method was implemented as such to mirror the mysql connection creation method. I agree both that it could be changed so it does not catch the error and return None, instead propagating it to the calling function; and that it probably isn't worth doing this as the functionality works either way, and would mean more code changes. I'm going to leave this as is for now

tests/test_db.py
Chris-Friend: See response to above related comment

crawler/db.py
Chris-Friend: I've made (and tested) these changes and will push up once I've addressed all comments

crawler/file_processing.py
Chris-Friend: I think @pjvv is doing some code cleaning in the next branch. Might be worth including this in that clean up?

crawler/helpers.py
Chris-Friend: Yeah this is the standard way in which known errors are added, to be referred to by TYPE in the calling code

tests/test_db.py
Chris-Friend: Yeah this looks like it should not raise an exception but should instead return `None`. Will investigate

Chris-Friend: The try/except was catching an `Error` where I think it should have been catching an `Exception`. Will check which was desired and change as appropriate

Chris-Friend: Updated to explicitly catch mysql.Errors and pyodbc.Errors where appropriate

crawler/file_processing.py
pjvv: I've done some linting and black auto formatting in branch "maintenance" but will leave refactoring to another branch to keep things sane.

pjvv (approved):

PR #151, opened 2020-11-11
Maintenance

pjvv:
Closes #

*
*
* ...

### DIFF
--- a/.github/workflows/ci_docker.yml
+++ b/.github/workflows/ci_docker.yml
@@ -1,7 +1,7 @@
-name: Docker CI
+name: CI docker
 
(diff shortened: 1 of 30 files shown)


Chris-Friend (approved):
I haven't looked over all the style changes in huge amounts of detail, but I like the CI changes. When exactly will the code style, format, quality etc. be analysed?

pjvv:
Great, thanks, we can refine things as we go but at least the baseline is set now.

It runs at every push to the develop and master branches and pull requests to these branches.

I'd like to optimise it to only run the docker CI when we need the image.

PR #160, opened 2020-11-26
GPL-733 dart migration

### LINKED ISSUE #144: GPL-733 As a developer I would like to run a Manual MLWH and DART Migration
**User story**
As developers we want to ensure that if a sample upload to the MLWH or DART databases fails with a critical exception, but the upload to mongo succeeds, we are able to run a manual process to pull samples from mongodb added between timestamps and attempt to re-add them to MLWH and DART (see existing manual MLWH migration that already exists in crawler - we want a combined version that keeps both in sync). The process needs to be idempotent, such that data that already exists is not overwritten. It will help to ensure the mongo, MLWH and DART databases remain in-sync  through MLWH and DART upload failures. NB. DART inserts should only be done if samples for a particular plate have NOT already been cherrypicked. **This does not handle changing filtered positive rules. See [GPL-709](https://github.com/sanger/crawler/issues/143) for this** 

**Acceptance criteria**
To be considered successful the solution must:
- Add a new combined MLWH/DART migration or update the existing MLWH migration that:
  - [x] requires start and end timestamp input parameters
  - [x] performs the existing MLWH migration functionality
  - [x] for DART fetches all RESULT=Positive samples from mongo between the timestamps. Also check for `filtered_positive_version`, `lh_sample_uuid` and `lh_source_plate_uuid` properties. If not we shouldn't add them as DART would become inconsistent/Beckman workflow expects these properties
  - [x] for DART, further filters these above samples on which aren't cherrypicked (if any samples in a plate are cherrypicked, don't add the plate) E.g. by checking for the sample cherrypicked event in MLWH (see get_cherrypicked_samples method in lighthouse); or be checking the MLWH lighthouse_sample table for a COG UK Id (as currently samples only get a COG UK Id when selected for cherrypicking - may be susceptible to future changes)
  - [x] ~~**Needs discussing**: Re-determine if filtered positive? _Is this required?_ Filtered positive fields should already be recorded in mongo, and there is a separate migration to update databases following filtered positive rule changes ([GPL-709](https://github.com/sanger/crawler/issues/143))~~
  - [x] ~~**Needs discussing**:  Re-add to Mongo & MLWH (_only necessary if re-determining filtered positive fields_)~~
  - [x] Add samples to DART (plates and well properties - see [GPL-709](https://github.com/sanger/crawler/issues/143)'s implementation and [GPL-745](https://github.com/sanger/crawler/pull/158))
- [x] include README.md documentation of this new process
- [x] ~~in the normal file processing, don't attempt to add to DART if the MLWH insert fails~~ Split to https://github.com/sanger/crawler/issues/162

**Dependencies**
- [GPL-708](https://github.com/sanger/crawler/issues/145)
- [GPL-709](https://github.com/sanger/crawler/issues/143)

**Additional context**
NB. There are specific error types that will show for a file in the Lighthouse-UI imports screen. This will need to be monitored to indicate to us when we need to run the migration.
The migration should also be able to insert legacy data for us, e.g. to get the initial plates into DART for cherrypicking.

pjvv:
Closes #144


migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+            print(f"Updating MLWH database for {len(mongo_docs_for_sql)} sample documents")
+            # create connection to the MLWH database
+            with create_mysql_connection(config, False) as mlwh_conn:
+
+                # execute sql query to insert/update timestamps into MLWH
+                run_mysql_executemany_query(mlwh_conn, SQL_MLWH_MULTIPLE_INSERT, mongo_docs_for_sql)
+        else:
+            print("No documents found for this timestamp range, nothing to insert or update in MLWH")
+
+        # 7. add all plates to DART, but only RESULT=Positive samples
```
emrojo, line 180: But all samples are Result=Positive in this method, or am I wrong?

emrojo (approved):
All good

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+            )
+
+            # 3. add the relevant filtered positive fields if not present
+            filtered_positive_identifier = FilteredPositiveIdentifier()
+            version = filtered_positive_identifier.current_version()
+            update_timestamp = datetime.now()
+
+            print("Updating filtered positives...")
+
+            update_filtered_positive_fields(
```
Chris-Friend, line 145: This will update the filtered positive fields on **all samples**. This should be fine as long as any pending filtered positive rule change migration is run first. If not, one situation in which it might not be okay is for a sample currently being cherrypicked, but not actually picked: you could update this sample to be not eligible for cherrypicking in mongo and MLWH, but this wouldn't go to DART as the sample's plate would be 'started'.

The safest way to do this would be to add filtered positive info if none existed on the sample (e.g. one of the three expected fields doesn't exist), as you know that sample won't be in the process of being cherrypicked. Although I can't think of a non-rare set of circumstances in which this will manifest as a problem, and the one I can think of isn't that much of a problem (picking a sample that has just been updated so it shouldn't be picked).

Will let you decide whether mitigating a small, possibly insignificant issue is worth adding more complexity here. Maybe just a comment?

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+                FIELD_FILTERED_POSITIVE: False,
+                FIELD_FILTERED_POSITIVE_VERSION: version,
+                FIELD_FILTERED_POSITIVE_TIMESTAMP: update_timestamp,
+            }
+        },
+    )
+    return True
+
+
+def samples_updated_with_source_plate_uuids(mongo_db, samples: List[Sample]) -> List[Sample]:
```
Chris-Friend, line 238: Would it be difficult to pull this logic shared with the `file_processing.py` equivalent out to a shared location? The different types or logging/error reporting may make this a bit fiddly

migrations/helpers/dart_samples_update_helper.py
Chris-Friend: This is correct. I think Phil is just mapping back to the above list of things todo. Might be worth reminding the reader that if the method gets here, all samples we pass through to this method will be positive?

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+def add_sample_uuid_field(samples: List[Sample]) -> List[Sample]:
+    # add lh sample uuid
+    for sample in samples:
+        if FIELD_LH_SAMPLE_UUID not in [*sample]:
+            sample[FIELD_LH_SAMPLE_UUID] = str(uuid.uuid4())
+
+    return samples
+
+
+def update_mongo_fields(mongo_db, samples: List[Sample], version: str, update_timestamp: datetime) -> bool:
```
Chris-Friend, line 195: Looks like you've borrowed some logic from the filtered positive migration. Worth extracting out to a shared location?

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+
+            # 4. add the UUID fields if not present
+            add_sample_uuid_field(samples)
+
+            samples_updated_with_source_plate_uuids(mongo_db, samples)
+
+            # 5. update samples in mongo updated in either of the above two steps (would expect the same set of samples
+            #       from both steps)
+            print("Updating Mongo...")
+            _ = update_mongo_fields(mongo_db, samples, version, update_timestamp)
```
Chris-Friend, line 162: Would it not be best to do a "if mongo fails, don't update MLWH or DART; if MLWH fails, don't update DART" workflow? Or am I missing that this is already in place (with exceptions?)?

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+        for sample in samples:
+            mongo_docs_for_sql.append(map_mongo_doc_to_sql_columns(sample))
+
+        if number_docs_found > 0:
+            print(f"Updating MLWH database for {len(mongo_docs_for_sql)} sample documents")
+            # create connection to the MLWH database
+            with create_mysql_connection(config, False) as mlwh_conn:
+
+                # execute sql query to insert/update timestamps into MLWH
+                run_mysql_executemany_query(mlwh_conn, SQL_MLWH_MULTIPLE_INSERT, mongo_docs_for_sql)
```
Chris-Friend, line 176: Do we need to update the SQL so on duplicate key it updates the fields we've added, specifically sample uuid, source plate uuid and the three filtered positive fields?

Chris-Friend: In fact as we're not adding anything new, it might be worth adding something like the `SQL_MLWH_MULTIPLE_FILTERED_POSITIVE_UPDATE` query

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+        else:
+            print("No documents found for this timestamp range, nothing to insert or update in MLWH")
+
+        # 7. add all plates to DART, but only RESULT=Positive samples
+        update_dart_filtered_positive_fields(config, samples)
```
Chris-Friend, line 181: Might be worth renaming this method where it is defined, as it doesn't just update the filtered positive fields, it updates everything

README.md
```diff
@@ -8,17 +8,24 @@
...
-- [Mypy](#mypy)
-- [Miscellaneous](#miscellaneous)
-  - [Naming conventions](#naming-conventions)
+* [Requirements](#requirements)
+* [Running](#running)
+* [Migrations](#migrations)
+  * [Updating the MLWH lighthouse_sample table](#updating-the-mlwh-lighthouse_sample-table)
+  * [Filtered Positive Rules](#filtered-positive-rules)
+    * [Version 1 `v1` - **Current Version**](#version-1-v1---current-version)
+    * [Propagating Filtered Positive version changes to MongoDB, MLWH and DART](#propagating-filtered-positive-version-changes-to-mongodb-mlwh-and-dart)
```
Chris-Friend, line 21: Missing documentation on new migration? Also worth removing the MLWH-only one, if the new migration is going to replace it?

Chris-Friend (changes requested):
Some good stuff here. A few comments about the specifics of the workflow. Also, looks like there aren't any tests for the new migration?

crawler/helpers/logging_helpers.py
```diff
@@ -0,0 +1,383 @@
...
+# See confluence for full table of aggregate types
+# https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/i.+Low+Occupancy+Cherry+Picking
+
+
+class AggregateType1(AggregateTypeBase):
+    def __init__(self):
+        super().__init__()
+        self.type_str = "TYPE 1"
+        self.error_level = ErrorLevel.DEBUG
+        self.message = f"{self.error_level.name}: Blank rows in files. ({self.type_str})"
```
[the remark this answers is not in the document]
pjvv: :D

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+                root_sample_ids.append(doc[FIELD_ROOT_SAMPLE_ID])
+                plate_barcodes.add(doc[FIELD_PLATE_BARCODE])
+
+                num_docs += 1
+
+            print(f"{num_docs} to process")
+            print(f"{len(plate_barcodes)} unique plate barcodes")
+
+            # 2. of these, find which have been cherry-picked and remove them from the list
+            cp_samples_df = get_cherrypicked_samples(config, root_sample_ids, plate_barcodes)
```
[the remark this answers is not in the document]
pjvv: Agree to both comments so made it a set.

migrations/helpers/dart_samples_update_helper.py
pjvv: Thanks, added more accurate comments.

crawler/file_processing.py
```diff
@@ -459,14 +446,17 @@ def get_db(self) -> Database:
...
+
         return db
 
-    # Database clash
-    def add_duplication_errors(self, exception):
+    def add_duplication_errors(self, exception: BulkWriteError) -> None:
+        """Database clash
+
+        Args:
+            exception ([type]): [description]
```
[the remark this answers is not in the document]
pjvv: Good spot, pushed a change to make consistent.

crawler/file_processing.py
```diff
@@ -568,20 +551,17 @@ def update_doc_from_source_plate(doc, existing_plate, skip_lab_check=False) -> N
...
-                "Error assigning source plate UUIDs to samples in file " f"{self.file_name}: {e}"
-            )
+            logger.critical("Error assigning source plate UUIDs to samples in file " f"{self.file_name}: {e}")
             logger.exception(e)
             updated_docs = []
 
         return updated_docs
 
-    def insert_samples_from_docs_into_mongo_db(self, docs_to_insert) -> List[ObjectId]:
+    def insert_samples_from_docs_into_mongo_db(self, docs_to_insert: List[Dict[str, str]]) -> List[Any]:
```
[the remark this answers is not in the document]
pjvv: In short, I don't think so because a Sample type is `Dict[str, Any]` but I think the "rows" from the CSV parser will be `Dict[str, str]`. I was still getting to grips with type aliases and figuring out where best to include the aliases for readability but didn't add a complete set of aliases yet.

crawler/file_processing.py
```diff
@@ -605,42 +585,35 @@ def insert_samples_from_docs_into_mongo_db(self, docs_to_insert) -> List[ObjectI
...
             return inserted_ids
         except Exception as e:
             logger.critical(f"Critical error in file {self.file_name}: {e}")
             logger.exception(e)
             return []
 
-    def insert_samples_from_docs_into_mlwh(self, docs_to_insert) -> None:
-        """Insert sample records into the MLWH database from the parsed file information, including
-        the corresponding mongodb _id
+    def insert_samples_from_docs_into_mlwh(self, docs_to_insert: List[Dict[str, str]]) -> None:
```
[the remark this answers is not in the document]
pjvv: Yip, we can revisit this when we have a little more time for a proper refactor too.

tests/test_helpers.py
```diff
@@ -30,25 +37,17 @@
...
-    DART_ROOT_SAMPLE_ID,
-    DART_RNA_ID,
-    DART_LAB_ID,
-    DART_LH_SAMPLE_UUID,
-    DART_STATE_PICKABLE,
-    DART_EMPTY_VALUE,
 )
-from crawler.helpers import (
-    LoggingCollection,
+from crawler.helpers.general_helpers import (
```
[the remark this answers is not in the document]
pjvv: Yes, good idea!

migrations/helpers/dart_samples_update_helper.py
pjvv: Yes, I need to implement that check, thanks for the reminder.

pjvv: Actually, any failure should throw an exception and stop but will add a few raises to double check my understanding.

migrations/helpers/dart_samples_update_helper.py
pjvv: Yes, that is a good idea, I think it's better to use a more specific query.

migrations/helpers/dart_samples_update_helper.py
pjvv: True, I've renamed it across the project.

README.md
pjvv: I've added the new one and thought of keeping the MLWH one until DART is up and running so that we have a fall-back in case we need it.

crawler/file_processing.py
```diff
@@ -394,9 +392,7 @@ def process_samples(self, add_to_dart: bool) -> None:
         if len(mongo_ids_of_inserted) > 0:
             # filter out docs which failed to insert into mongo - we don't want to create mlwh
             # records for these
-            docs_to_insert_mlwh = list(
-                filter(lambda x: x[FIELD_MONGODB_ID] in mongo_ids_of_inserted, docs_to_insert)
-            )
+            docs_to_insert_mlwh = list(filter(lambda x: x[FIELD_MONGODB_ID] in mongo_ids_of_inserted, docs_to_insert))
 
             self.insert_samples_from_docs_into_mlwh(docs_to_insert_mlwh)
```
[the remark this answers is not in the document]
pjvv: Yes, let's do this in another story.

run_migration.py
```diff
@@ -59,6 +69,7 @@ def migration_by_name(migration_name):
     switcher = {
         "sample_timestamps": migration_sample_timestamps,
         "update_mlwh_with_legacy_samples": migration_update_mlwh_with_legacy_samples,
```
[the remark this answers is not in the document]
pjvv: Yes I wanted to keep the existing one and then remove once DART is up and running. I think the MLWH updates should be idempotent now that I've used an update script instead.

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+        return
+
+    start_datetime = datetime.strptime(s_start_datetime, MONGO_DATETIME_FORMAT)
+    end_datetime = datetime.strptime(s_end_datetime, MONGO_DATETIME_FORMAT)
+
+    if start_datetime > end_datetime:
+        print("Aborting run: End datetime must be greater than Start datetime")
+        return
+
+    print(f"Starting DART update process with Start datetime {start_datetime} and End datetime " f"{end_datetime}")
```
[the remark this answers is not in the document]
pjvv: haha yip, typo!

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,387 @@
...
+            print(f"{num_docs} to process")
+            print(f"{len(plate_barcodes)} unique plate barcodes")
+
+            # 2. of these, find which have been cherry-picked and remove them from the list
+            cp_samples_df = get_cherrypicked_samples(config, root_sample_ids, plate_barcodes)
+
+            # create a second match for the pipeline
+            second_match = {
+                "$match": {
+                    FIELD_ROOT_SAMPLE_ID: {"$nin": cp_samples_df[FIELD_ROOT_SAMPLE_ID].to_list()},  # type: ignore
```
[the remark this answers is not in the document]
pjvv: Yes good point, I will revisit this.

migrations/helpers/dart_samples_update_helper.py
pjvv: Wow, yes good spot of an edge-case. I think I'll add more comments for now and revisit if time permits.

migrations/helpers/dart_samples_update_helper.py
pjvv: It probably wouldn't be too difficult but thought of leaving it for a larger refactor to move migrations "closer" to the main codebase. Currently it skirts the main `runner.py` but ideally we should be calling migrations using a sub-parser for argparse, i.e. `python runner.py migrate <name>`

crawler/file_processing.py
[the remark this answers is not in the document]
Chris-Friend: :+1:

README.md
```diff
@@ -74,56 +75,88 @@ The following runtime flags are available:
...
 
     python run_migration.py update_filtered_positives
 
-The process does not duplicate any data, instead updating existing entries.
+The process does not duplicate any data, instead updates existing entries.
+
+### Migrating legacy data to DART
+
+When the Beckman robots come online, we need to populate the DART database with the filtered positive samples that are
+available physically. This can be achieved using the 'update_dart' migration.
```
Chris-Friend, line 135: This can also be used similarly to the existing MLWH migration: if a DART insert process fails, you will see a critical exception for the file in the Lighthouse-UI. After addressing reason for failure, run between relevant timestamps to re-insert/update data into DART.

My original intention was to have this migration replace the existing MLWH one and do both MLWH and DART updates on failure, but I think we'd be best keeping the separate MLWH and DART migrations. What do you think? This will impact on the SQL query with which you update MLWH

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -147,13 +158,15 @@ def update_dart(config, s_start_datetime: str = "", s_end_datetime: str = "") ->
             # create connection to the MLWH database
             with create_mysql_connection(config, False) as mlwh_conn:
 
-                # execute sql query to insert/update timestamps into MLWH
-                run_mysql_executemany_query(mlwh_conn, SQL_MLWH_MULTIPLE_INSERT, mongo_docs_for_sql)
+                # execute SQL query to update filtered positive and UUID fields
+                run_mysql_executemany_query(
+                    mlwh_conn, SQL_MLWH_MULTIPLE_FILTERED_POSITIVE_AND_UUID_UPDATE, mongo_docs_for_sql
```
Chris-Friend, line 163: Partnering the README comment: this is fine if this migration won't be used to try to insert into MLWH. We'll have to keep the update MLWH and this migration to (re-)attempt MLWH inserts and then DART inserts, unless you fancy changing the SQL again? In which case you could have the insert SQL just update the UUID and filtered positive fields on duplicate key

Chris-Friend (commented):
Just a couple more comments about whether to merge the MLWH and DART migrations, and if so which MLWH SQL query to use

migrations/helpers/dart_samples_update_helper.py
Chris-Friend: As per your previous comment, resolving this on this PR, but still a valid option to extract out shared methods used both here and in the update filtered positives migration

migrations/helpers/dart_samples_update_helper.py
```diff
@@ -0,0 +1,392 @@
...
+            FIELD_CREATED_AT: {"$gte": start_datetime, "$lte": end_datetime},
+            FIELD_RESULT: {"$regex": "^positive", "$options": "i"},
+            FIELD_ROOT_SAMPLE_ID: {"$not": {"$regex": "^CBIQA_"}},
+        }
+    }
+    return list(samples_collection.aggregate([match]))
+
+
+def remove_cherrypicked_samples(samples: List[Sample], cherry_picked_ids: List[str]) -> List[Sample]:
+    return list(filter(lambda sample: sample[FIELD_ROOT_SAMPLE_ID] not in cherry_picked_ids, samples))
```
Chris-Friend, line 189: I still think that to remove cherrypicked samples you need to check on a (root sample id, plate barcode) tuple rather than just plate barcode. As multiple samples can exist with the same root sample id, with the potential for one being cherrypicked, and one not

tests/migrations/helpers/test_dart_samples_update_helper.py
```diff
@@ -0,0 +1,109 @@
...
+    source_plate = new_mongo_source_plate(plate_barcode, lab_id)
+
+    assert {FIELD_LH_SOURCE_PLATE_UUID, FIELD_BARCODE, FIELD_LAB_ID, FIELD_UPDATED_AT, FIELD_CREATED_AT} == set(
+        [*source_plate]
+    )
+    assert uuid.UUID(str(source_plate[FIELD_LH_SOURCE_PLATE_UUID]))
+    assert source_plate[FIELD_BARCODE] == plate_barcode
+    assert source_plate[FIELD_LAB_ID] == lab_id
+    assert source_plate[FIELD_UPDATED_AT] == now
+    assert source_plate[FIELD_CREATED_AT] == now
```
Chris-Friend, line 109: Will you be adding more tests to this, e.g. testing the `update_dart` method, and that failed inserts into dbs doesn't attempt to insert into subsequent dbs?

migrations/helpers/dart_samples_update_helper.py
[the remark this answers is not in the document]
Chris-Friend: :+1: left a new comment for this

migrations/helpers/dart_samples_update_helper.py
pjvv: Let's update the migration to use the enhancement required in #162 .

README.md
pjvv: I still think it's a good idea for this to replace the MLWH migration but let's do that once we're happy with how it performs with a live DART DB?

migrations/helpers/dart_samples_update_helper.py
pjvv: I don't mind changing the SQL. I think it's best to have it update on duplicate key to make it converge closer to what we want it to do when we deprecate the migrate MLWH migration.

README.md
```diff
@@ -63,57 +71,96 @@ The following runtime flags are available:
...
+
+This can also be used similarly to the existing MLWH migration: if a DART insert process fails, you will see a critical
+exception for the file in the Lighthouse-UI. After addressing reason for failure, run between relevant timestamps to
+re-insert/update data into DART.
```
Chris-Friend, line 139: "re-insert/update data into _MLWH and then_ DART"

migrations/helpers/dart_samples_update_helper.py
Chris-Friend: Looks like you've change to use the insert but update on duplicate key SQL :+1:

tests/migrations/helpers/test_dart_samples_update_helper.py
```diff
@@ -164,13 +210,104 @@ def test_update_dart_returns_early_invalid_end_datetime(mock_mongo_client, mock_
...
+
+        root_sample_ids, plate_barcodes = extract_required_cp_info(samples)
+
+        mock_mysql_connection.assert_not_called()
+        mock_get_cherrypicked_samples.assert_called_once_with(config, list(root_sample_ids), list(plate_barcodes))
+
+
+def test_update_dart_get_cherrypicked_samples_mocked_with_samples(
+    config, mongo_database, mock_mysql_connection, mock_update_dart
+):
```
emrojo, line 263: I would rename test nameto describe what is happening inside, like test_migrate_all_dbs_calls_remove_cherrypicked_samples

tests/migrations/helpers/test_dart_samples_update_helper.py
```diff
@@ -164,13 +210,104 @@ def test_update_dart_returns_early_invalid_end_datetime(mock_mongo_client, mock_
...
+
+            samples = get_positive_samples(mongo_db.samples, start_datetime, end_datetime)
+
+            mock_remove_cherrypicked_samples.assert_called_once_with(
+                samples, [[cherry_picked_sample[FIELD_ROOT_SAMPLE_ID][0], cherry_picked_sample[FIELD_PLATE_BARCODE][0]]]
+            )
+
+
+def test_update_dart_remove_cherrypicked_samples(config, mongo_database, mock_update_dart):
+    _, mongo_db = mongo_database
```
emrojo, line 298: Same here, test_ migrate_all_dbs_not_calling_remove_cp_samples

emrojo (approved):

PR #170, opened 2020-12-10
GPL-740 new fit to pick control rules

### LINKED ISSUE #167: GPL-771 Migrate all samples to use the new filtered_positive fields
**User story**

Katy T
Eduardo M
Phil
Chris F

**Acceptance criteria**

**Additional context**

### LINKED ISSUE #169: GPL-699 Result 'Limit of detection' == 'Positive'
**User story**
GPL-699 | As Lead Covid-19 Statistical Geneticist (Jeff B) I would like samples with a result of 'Limit of detection' to be treated the same as a positive result in col E of plate map files

Rich L
Cristina A

**Acceptance criteria**
- [ ] On finding value of 'Limit of detection' in result column E change crawler behaviour to see this as positive to be picked, given current Ct rules 

**Dependencies**
Need to clarify if values 'Detected' and 'Not Detected' are Positive and Negative respectively

**Additional context**
Data is parsed today as result of GPL-654. As of 22nd Oct, no results of 'Limit of detection' have been seen in wild.

harrietc52:
Closes #167 

* Updates filtered positive sample rules to V2

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,14 @@ in the below relevant section.
 The implementation of the current version can be found in [FilteredPositiveIdentifier](./crawler/filtered_positive_identifier.py),
 with the implementation of previous versions (if any) in the git history.
 
+#### Version 0 `v0`
+
+A sample is filtered positive if:
+
+* it has a positive RESULT
+
+This is the pre-"fit-to-pick" implementation, without any extra filtering on top of the RESULT=Positive requirement.
+
 #### Version 1 `v1`
 
 A sample is filtered positive if:
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Confluence page.
 
 A sample is filtered positive if:
 
-* it has a 'Positive' or 'limit of detection' RESULT
+* it has a 'Positive'
 * it is not a control (ROOT_SAMPLE_ID does not start with 'CBIQA_', 'QC0', or 'ZZA000')
 * all of CH1_CQ, CH2_CQ and CH3_CQ are `None`, or one of these is less than or equal to 30
 
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Confluence page.
 
 A sample is filtered positive if:
 
-* it has a 'Positive'
+* it has a 'Positive' RESULT
 * it is not a control (ROOT_SAMPLE_ID does not start with 'CBIQA_', 'QC0', or 'ZZA000')
 * all of CH1_CQ, CH2_CQ and CH3_CQ are `None`, or one of these is less than or equal to 30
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -115,6 +123,17 @@ A sample is filtered positive if:
 More information on this version can be found on [this](https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/UAT+6th+October+2020)
 Confluence page.
 
+#### Version 2 `v2` - **Current Version**
+
+A sample is filtered positive if:
+
+* it has a 'Positive'
```
Chris-Friend, line 130: * it has a 'Positive' **RESULT**

crawler/filtered_positive_identifier.py
```diff
@@ -44,23 +42,58 @@ def is_positive(self, sample: Sample) -> bool:
...
+
+def current_filtered_positive_identifier() -> FilteredPositiveIdentifier:
+    return FilteredPositiveIdentifierV2()
+
+
+class FilteredPositiveIdentifierV0(FilteredPositiveIdentifier):
+    def __init__(self):
+        super(FilteredPositiveIdentifierV0, self).__init__()
+        self.version = FILTERED_POSITIVE_VERSION_0
+        self.result_regex = re.compile(f"^{POSITIVE_RESULT_VALUE}", re.IGNORECASE)
```
Chris-Friend, line 81: The `result_regex` is the same for each version. Could be extracted out to the parent class

Chris-Friend (approved):

crawler/filtered_positive_identifier.py
```diff
@@ -14,23 +15,20 @@
...
-        """Returns the current version of the identifier.
 
-        Returns:
-            {str} - the version number
-        """
-        return self.versions[-1]
+class FilteredPositiveIdentifier(ABC):
+    def __init__(self):
+        self.version = None
+        self.ct_value_limit = decimal.Decimal(30)
```
emrojo, line 27: Doesn't it need ```create_decimal128_context``` to create this decimal?

emrojo (approved):

crawler/filtered_positive_identifier.py
Chris-Friend: Hmmm not sure. I have made the change to define within a context, and the tests still pass, so will go with that

Chris-Friend: Actually, looking at what we do I don't think so. This value is just a `Decimal`. Then later on we, in the `d128_context` covert `Decimal128` values to `Decimal` to allow comparing. I don't think defining the ct value limit `Decimal` need be done in the context

PR #171, opened 2020-12-10
GPL-757 hotifx - Files with wrong format break the crawler process

KatyTaylor:
Branched from master and going back into master to bypass the Beckman stuff.

Afterwards, I will merge this back into develop to keep it up to date. I have already checked the conflicts and it's not bad.

**Commits:**

add error handling for scenario where xlsx file is processed, with csv extension

- add new test in test_file_processing
- had to modify numbers in test_main to keep them passing - rethink whether can add a test so don't need to change these


tests/test_file_processing.py
```diff
@@ -106,6 +106,25 @@ def test_process_files(mongo_database, config, testing_files_for_process, testin
...
+    centre_config["sftp_root_read"] = "tmp/files"
+    centre = Centre(config, centre_config)
+    centre.process_files()
+
+    imports_collection = get_mongo_collection(mongo_database, COLLECTION_IMPORTS)
+    samples_collection = get_mongo_collection(mongo_database, COLLECTION_SAMPLES)
+    samples_history_collection = get_mongo_collection(mongo_database, COLLECTION_SAMPLES_HISTORY)
+
+    # check that the valid file still gets processed, even though the bad file is in there
+    assert samples_collection.count_documents({"RNA ID": "789_A02", "source": "Test Centre"}) == 1
```
Chris-Friend, line 126: Could you also check that the appropriate error has been logged? Something like:
```
assert centre_file.logging_collection.get_count_of_all_errors_and_criticals() == 1
assert centre_file.logging_collection.aggregator_types["TYPE 10"].count_errors == 1
```

Chris-Friend (approved):
Decided to jump on reviewing this 🙂. Just one minor comment

tests/test_file_processing.py
KatyTaylor: Thanks, yes will do

KatyTaylor: Done - didn't have access to the CentreFile instance, but can check the imports collection messages.

Chris-Friend (dismissed):

Issue #173, opened 2020-12-11
GPL-nnn Refactor the update_dart migration


Chris-Friend:
**User story**
As developers we want to easily be able to understand, update and test different pieces of functionality. This is made considerably more difficult if lots of code is placed in a single file. We should aim to separate logically similar functionality into separate files. This way we can use mocking to reduce test complexity, but also group together common functionality in shared helper files.

In this instance it makes sense to separate into different files script-type logic from the methods it calls, so we can easily, separately test:
- the script logic
- the methods called at each step in the script do the job we expect of them

**Who are the primary contacts for this story**
Chris F

**Steps to Aid Completion**
- [ ] Have the `update_dart.py` migration log instead of print
- [ ] Move the implementation of `migrate_all_dbs` in `dart_samples_update_helper.py` to the `run` method of `update_dart.py`. Keep all other methods in the helper file
- [ ] Create corresponding test file, moving any relevant tests from `test_dart_samples_update_helper.py`
- [ ] Add more tests to the new test file, including (mocked) end-to-end tests
- [ ] Add more tests for the helper methods left in `dart_samples_update_helper.py`
- [ ] Do some integration testing: either in the development environment; or by adding some tests that have less mocking (e.g. do not mock the databases)

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] Separation of script-type logic in the update_dart migration from the helper methods called as part of the script 

**Dependencies**
This story is blocked by the following dependencies:
- https://github.com/sanger/crawler/issues/168

**Additional context**
As the migration wasn't/isn't "fully" tested, some issues were found when updating it (above linked story), and more still may remain

Issue #150, opened 2020-11-11
GPL-740 Filter our Glasgow controls from pick-able positives


rl15:
**User story**
GPL-740 As POC for LightHouse LAbs for PAM (Alan K) I would like the Glasgow controls to be ignored from pick even if they are positive with low Cts as their sequence is of no value for outbreak investigation, surveillance or vaccine trials

**Who are the primary contacts for this story**
Alan K (PAM)
Rich L

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] Discard samples with root ids starting 'ZZA000' as positives to pick
- [x] Discard samples with root ids starting 'QC0' as positives to pick (NB this is a zero)

**Additional context**
See GPL-648 Filter out LH Cambridge internal control samples #129 (in production ~16th Oct)

rl15:
Alan K wrote (Wednesday, 11 November 2020 at 14:36)

I am happy with these acceptance criteria.

PR #174, opened 2020-12-11
Gpl-775 legacy samples migration update

### LINKED ISSUE #168: GPL-775 Do not update filtered positive values on running migration to add legacy data to DART
**User story**
In the #167 migration we require that all samples are given fit to pick rules. It is the intention that this migration is run once, before all other migrations, to ensure that all samples (old and new) have filtered positive values. As such, the #144 migration to load legacy data into DART needs to be updated to not update filtered positive values. 

Philip
Chris F

**Acceptance criteria**
- [x] update the migration to load legacy data into DART to not update filtered positive values, and therefore not update mongo/mlwh with any new filtered positive values. Note that new sample and source plate uuid values will still be required

**Additional context**

Chris-Friend:
Closes #168 

* Do not update filtered positive fields as part of the "making legacy samples ready for the Beckman process" migration
* Do not exclude control samples from the migration
* Update mongo sample's UUIDs (this was missed from the original implementation)
* Add more tests to the migration
* Spawning of the related task https://github.com/sanger/crawler/issues/173 to separate code that has different purposes into different files


migrations/helpers/dart_samples_update_helper.py
```diff
@@ -202,7 +168,6 @@ def get_positive_samples(
             # 1. First filter by the start and end dates
             FIELD_CREATED_AT: {"$gte": start_datetime, "$lte": end_datetime},
             FIELD_RESULT: {"$regex": "^positive", "$options": "i"},
-            FIELD_ROOT_SAMPLE_ID: {"$not": {"$regex": "^CBIQA_"}},
```
pjvv, line 205: Why remove the control check?

Chris-Friend: To avoid duplication of the control definitions in multiple places: we only want this changing definition to be a part of the fit to pick rules. It does not matter if we add information on a control to DART/for the Beckman process. We know that the fit to pick rules will ensure this is not picked

Chris-Friend: e.g. https://github.com/sanger/crawler/issues/150 adds more control rules, which we'd also have to add to here if we wanted to filter out controls from these samples

PR #178, opened 2020-12-15
Gpl 765 update fit to pick migration

### LINKED ISSUE #163: GPL-765 Update the fit to pick rule change migration to not update cherrypicked samples
**User story**
We do not want to update the filtered positive information for cherrypicked samples when running the filtered positive rule change migration.

The migration currently updates Result=Positive samples in 'pending' DART plates. But on adding legacy data as #144 will allow, we introduce the possibility of adding plates to DART that have cherrypicked samples on them. Note we don't add the cherrypicked samples to DART, just the plates, but the migration workflow doesn't currently account for plates in DART that have cherrypicked samples on them:

* On day one, all Result=Positive non-cherrypicked samples are added to DART as part of the legacy data migration (#144 )
* Some time later, the filtered positive rules change, and a filtered positive migration is required. This gets all pending plates from DART, then **all Result=Positive samples** in these plates from Mongo. This will include samples that have already been cherrypicked 😧. So we end up updating the filtered positive rule changes on samples that have already been picked and should remain unchanged 💥

_**Up for discussion**: why do we need the "get pending plates from DART" step and the non-cherrypicked samples step? Would it be sufficient to just update non-cherrypicked samples, as they should be on pending DART plates anyway after the initial migration: non-cherrypicked samples_ should be _a subset of samples on pending DART plates (e.g. if a plate is processed using the Sentinel workflow, and the source plate left pending in DART)? Is there harm in being rigorous and including both steps?_

Chris F
Phil

**Acceptance criteria**
- [x] #144 and https://github.com/sanger/General-Backlog-Items/issues/93 should introduce to crawler a way of determining if a sample is cherrypicked or not. Use this to add an extra step to the filtered positive rule change migration to not update cherrypicked samples
- [x] Allow omitting DART from the migration, hence allowing it to be run idependently of the Beckman process
- [ ] ~~(**See discussion point above**) Potentially remove the "sample has to be on a pending DART plate" check in favour of just the not cherrypicked one~~ See comment for decision

**Dependencies**
- #144
- https://github.com/sanger/General-Backlog-Items/issues/93 - will include changes to the same code

**Additional context**

Chris-Friend:
Closes #163 

* Extract some common migration functionality to the `shared_helper.py` file
* Filter out cherrypicked samples when updating fit-to-pick rules


migrations/update_filtered_positives.py
```diff
@@ -39,41 +40,56 @@ def run(settings_module: str = "") -> None:
...
+                        config, non_cp_pos_pending_samples, version, update_timestamp
+                    )
+                    logger.info("Finished updating Mongo")
+
+                    if mongo_updated:
+                        logger.info("Updating MLWH...")
+                        mlwh_updated = update_mlwh_filtered_positive_fields(config, non_cp_pos_pending_samples)
+                        logger.info("Finished updating MLWH")
+
+                        if mlwh_updated:
```
pjvv, line 85: Should we add a check or setting in the event that we don't go with Beckman but still want to run this migration?

Chris-Friend: I think this is something we should be able to do. We'd just not get pending plate barcodes from DART, instead getting all RESULT=Positive samples from mongo, and not update DART at the end. Do you think it worth it? It'd be a little extra work, but not too much

pjvv: Tough call. They could delay the decision on Beckman till next year sometime but we might want to run this sooner? We have some time while we wait on unified_warehouse etc. so we can see if there is any news tomorrow morning?

Chris-Friend: I think it worthwhile as it allows more separation of the Beckman workflow and the fit-to-pick rules. My intention would be to implement it opposite to the normal crawler, i.e. attempt to query DART unless told not to

Chris-Friend: Done

migrations/update_filtered_positives.py
```diff
@@ -37,73 +38,80 @@ def run(settings_module: str = "") -> None:
...
+                    logger.info("Finished updating MLWH")
+
+                    if not omit_dart and mlwh_updated:
+                        logger.info("Updating DART...")
+                        dart_updated = update_dart_fields(config, non_cp_pos_pending_samples)
+                        logger.info("Finished updating DART")
             else:
-                logger.warning("No positive samples in pending plates found in Mongo, not updating any database")
+                logger.warning(
+                    "No non-cherrypicked positive samples in pending plates found, not updating any database"
```
pjvv, line 100: This message is a little misleading as it does not always look in pending plates.

migrations/update_filtered_positives.py
```diff
@@ -37,73 +38,80 @@ def run(settings_module: str = "") -> None:
...
+                logger.info(f"{num_pending_plates} pending plates found in DART")
+
+                # Get positive result samples from Mongo in these pending plates
+                logger.info("Selecting postive samples in pending plates from Mongo...")
+                samples = positive_result_samples_from_mongo(config, pending_plate_barcodes)
+            else:
+                logger.warning("No pending plates found in DART")
+
+        if num_samples := len(samples):
+            logger.info(f"{num_samples} positive samples in pending plates found in Mongo")
```
pjvv, line 61: This message is a little misleading as it does not always look in pending plates.

migrations/update_filtered_positives.py
```diff
@@ -37,73 +38,80 @@ def run(settings_module: str = "") -> None:
...
+
+            # Filter out cherrypicked samples
+            logger.info("Filtering out cherrypicked samples...")
+            non_cp_pos_pending_samples = remove_cherrypicked_samples(config, samples)
```
pjvv, line 65: The variable naming might be confusing too with pending but this is only if we don't go ahead with DART. If we do, then we will always want to update DART too.

pjvv: So might not be worth the refactor now.

pjvv (approved):

migrations/update_filtered_positives.py
Chris-Friend: I agree. Have updated to not assume DART is included in the migration

migrations/update_filtered_positives.py
Chris-Friend: Always worth a clarity refactor 😄

migrations/helpers/shared_helper.py
```diff
@@ -1,8 +1,23 @@
+from datetime import datetime
```
KatyTaylor, line 1: I would make the name of this file less generic, as it is only relevant to the update_filtered_positives migration, not all migrations. Wouldn't want someone to put some completely unrelated code in in future.

KatyTaylor: Ah I see it's already got some shared stuff in it. In which case I would make a new file to put these in

Chris-Friend: The methods that I extracted are used in the `dart_samples_update_helper` and the `update_filtered_positives_helper`: they are no less shared than the methods that already existed in the file. It was my intention to use this file for the refactoring of the migrations https://github.com/sanger/crawler/issues/173

KatyTaylor (approved):

PR #183, opened 2020-12-16
GPL-749 space and single quotes

### LINKED ISSUE #153: GPL-749 BUG - Crawler accepting sample rows with spaces and single quotes in rna_id field
**Describe the bug**
Crawler is allowing rows to be processed that contain spaces and single quotes in the rna_id field. These then get through into the plate barcode field in MLWH (and presumably MongoDB) causing problems in reporting.
Also need to correct existing rows with this issue already in databases.

**RT Ticket Number**
If applicable

**To Reproduce**
Can be seen in files and databases

**Expected behaviour**
We should either strip these extra characters off and insert the corrected values, or reject the rows (and add one or more new error types in crawler).

**Additional context**
From Neil Friday 13th in slack:

> Seeing plate barcodes with leading/trailing back ticks in them on the plate maps, they get imported into mlwh but Limber/Events tables don’t making joins a bit tricky
eg
\`AP-rna-00020028 in plate maps plate_barcode
\`AP-rna-00020028 in mlwh.lighthouse_sample
AP-rna-00020028 in LabWhere.labware.barcode
AP-rna-00020028 in mlwh_events.subjects.friendly_name
Affects 2873 rows in lighthouse_sample (273 +ves)
There are also ~2K of samples whose plate barcodes have leading or trailing whitespace which can also perpetuate in LabWhere/Events.. or not
‘ AP-rna-00115445’ present as ‘AP-rna-00115445’ whilst ' GLA019659R’ is.
Would be good to have a fixed rule of clense or no clense

pjvv:
Closes #153 

* update regex which matches plate barcode and coordinate
* refactor settings to make testing more reliable


crawler/config/defaults.py
```diff
@@ -28,53 +28,56 @@
 # sftp_root_write: directory on sftp in which to upload master files
 # file_names_to_ignore: array of files to exclude from processing, such as those
 #                       containing invalid headers
+CENTRE_REGEX_BARCODE = r"^[\W_]*([\w-]*)_([a-z]\d{2})[\W_]*$"
+CENTRE_DIR_BACKUPS = "data/backups"
+CENTRE_REGEX_SFTP_FILE = r"sanger_report_(\d{6}_\d{4}).*\.csv$"
```
Chris-Friend, line 33: Looks like you can use this variable in all the centres, not just the first one

crawler/config/integration_with_blacklist_change.py
```diff
@@ -1,59 +1,12 @@
...
-        "backups_folder": "tmp/backups/CAMC",
-        "sftp_file_regex": r"^CB_sanger_report_(\d{6}_\d{4})\.csv$",
-        "sftp_root_read": "project-heron_cambridge-az",
-        "file_names_to_ignore": [],
-        "biomek_labware_class": "Bio-Rad_96PCR",
-    },
-]
+# test with no black-listed files
+for centre in CENTRES:
+    centre["file_names_to_ignore"] = []
```
Chris-Friend, line 12: Will this work? I.e. not re-defining the CENTRES config, but instead overriding values in the test config?

crawler/file_processing.py
```diff
@@ -983,31 +976,28 @@ def parse_and_format_row(self, row, line_number, seen_rows) -> Any:
...
                 modified_row[FIELD_PLATE_BARCODE],
                 modified_row[FIELD_COORDINATE],
             ) = self.extract_plate_barcode_and_coordinate(modified_row, line_number, barcode_field, barcode_regex)
+
         if not modified_row.get(FIELD_PLATE_BARCODE):
             return None
 
-        # add file details and import timestamp
-        import_timestamp = self.get_now_timestamp()
-
```
Chris-Friend, line 1000: Why remove this? It will add to the processing time to have to re-determine the current datetime for each of the three fields below. And there is no obvious reason we wouldn't expect these to be (very nearly) the same

crawler/file_processing.py
```diff
@@ -1327,11 +1317,10 @@ def new_mongo_source_plate(self, plate_barcode: str, lab_id: str) -> SourcePlate
         Returns:
             SourcePlate -- The new mongo source plate doc.
         """
-        timestamp = self.get_now_timestamp()
```
Chris-Friend, line 1330: Same here as for the other `timestamp` comment

crawler/file_processing.py
```diff
@@ -833,7 +827,7 @@ def log_adding_default_lab_id(self, row, line_number):
         )
 
     def filtered_row(self, row, line_number) -> Dict[str, Any]:
-        """Filter unneeded columns and add lab id if not present and config flag set.
+        """Filter un-needed columns and add lab id if not present and config flag set.
```
Chris-Friend, line 830: I think **_unneeded_** is the correct spelling

crawler/file_processing.py
```diff
@@ -132,20 +132,18 @@ def get_files_in_download_dir(self) -> List[str]:
...
-        """Remove the files downloaded from the SFTP for the given centre.
-
-        Arguments:
-            centre {Dict[str, str]} -- the centre in question
-        """
+        """Remove the files downloaded from the SFTP for the given centre."""
         logger.debug("Remove files")
         try:
             shutil.rmtree(self.get_download_dir())
+            (_, subfolders, files) = next(os.walk("tmp/files/"))
```
Chris-Friend, line 139: What does this do? Find the next temporary file? Why is it needed now and not before?

crawler/file_processing.py
```diff
@@ -426,11 +422,10 @@ def backup_file(self) -> None:
         shutil.copyfile(self.full_path_to_file(), destination)
 
     def create_import_record_for_file(self) -> None:
-        logger.info(f"{self.docs_inserted} documents inserted")
```
Chris-Friend, line 429: Do we no longer want to log in this method?

crawler/file_processing.py
```diff
@@ -783,25 +778,27 @@ def check_for_required_headers(self, csvreader: DictReader) -> bool:
...
-            barcode_field {str} -- field indicating the plate barcode of interest, might also
-            include coordinate
+            barcode_field {str} -- field indicating the plate barcode of interest, might also include coordinate
             regex {str} -- regex pattern to use to extract the fields
 
         Returns:
             Tuple[str, str] -- the barcode and coordinate
         """
-        m = re.match(regex, row[barcode_field])
+        pattern = re.compile(regex, re.IGNORECASE | re.ASCII)
```
Chris-Friend, line 796: What is the significance of defining the pattern with `re.IGNORECASE` and `re.ASCII`?

tests/test_file_processing.py
```diff
@@ -264,104 +263,116 @@ def test_row_required_fields_present(config):
...
+                "line_number": 2,
+                "Result": "Positive",
+                "file_name": "some file",
+                "file_name_date": None,
+                "created_at": now,
+                "updated_at": now,
+                "Lab ID": None,
+                "filtered_positive": True,
+                "filtered_positive_version": "v2.3",
+                "filtered_positive_timestamp": now,
```
Chris-Friend, line 345: This makes the tests more brittle: you expect the two separate calls to `datetime.now()` to be the same, but in reality finite amount of time passes between the two calls, hence they could return different values

As below for other changes

Chris-Friend (commented):
Just a couple of minor things

crawler/config/defaults.py
pjvv: Thanks, yeah I was testing if it would work and forgot to add it to the rest.

crawler/config/integration_with_blacklist_change.py
pjvv: I think it is working but I was having some problems with tests so it might be this. My understanding is that this should.

crawler/file_processing.py
pjvv: 🤣

crawler/file_processing.py
pjvv: This snuck in while debugging...

crawler/file_processing.py
pjvv: At first I didn't think it was too helpful but with more context I think it will be.

crawler/file_processing.py
pjvv: I just double checked and the IGNORECASE flag isn't required, by defining ASCII it covers upper and lower case.

tests/test_file_processing.py
pjvv: It's using the [freezegun](https://pypi.org/project/pytest-freezegun/) plugin though, to freeze time.

Chris-Friend: Ah. That makes sense

crawler/file_processing.py
Chris-Friend: What about for coordinates, e.g. `a01` vs. `A01`?

crawler/file_processing.py
pjvv: I'm going to move this logging to somewhere more appropriate, in the function that gets called.

crawler/file_processing.py
pjvv: I'm thinking hard why I did... 🤔  I think, it seems like adding extra complexity to just wrap a one liner standard library call (`datetime.now()`) and they seem to be used in sequential lines so the time difference is negligible. In terms of overhead, I don't think 2 extra calls would would make a noticeable difference?

crawler/file_processing.py
pjvv: Are you asking whether we should allow both?

crawler/file_processing.py
Chris-Friend: I'm happy to keep the change now you've added it, especially with the freezegun plugin testing solution

crawler/file_processing.py
Chris-Friend: ... yes, maybe. Specifically whether we want to allow lowercase coordinates, or whether to use `[A-Z]` and enforce uppercase like the previous implementation

pjvv: You're right, it's best to keep it as close to current implementation until they notice something and would like to allow lowercase. I'll update the regex and flags.

Chris-Friend (dismissed):

Chris-Friend (approved):

PR #185, opened 2020-12-17
Gpl 771 migrate legacy filtered positive

### LINKED ISSUE #167: GPL-771 Migrate all samples to use the new filtered_positive fields

alexperrin1:
Closes #167 

- Add migration to set the filtered positive fields on all legacy samples


migrations/helpers/update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,191 @@
...
+def v0_version_set(config: ModuleType):
+    """Find if the v0 or v1 version has been set in any of the samples.
+       This would indicate that the legacy migration has already been run.
+
+    Args:
+        samples {List[Dict]} -- List of samples from Mongo
+
+    Returns:
+        Boolean -- v0 version set in samples
+    """
```
Chris-Friend, line 141: Some typing/description issues here
* method should return a bool (declaration and description)
* description is not accurate: no longer looking for v1 samples
* arguments are wrong

migrations/helpers/update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,191 @@
...
+    with create_mongo_client(config) as client:
+        mongo_db = get_mongo_db(config, client)
+        samples_collection = get_mongo_collection(mongo_db, COLLECTION_SAMPLES)
+
+        v0_samples = list(
+            samples_collection.find({FIELD_FILTERED_POSITIVE_VERSION: FILTERED_POSITIVE_VERSION_0})
+        )  # noqa: E501
```
Chris-Friend, line 148: I think you can use `count_documents` with the same query to return the number directly, rather than pulling all the v0 samples into a memory. And just `return num_v0_samples > 0`, rather than the if-statement

migrations/helpers/update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,191 @@
...
+        if len(v0_samples) > 0:
+            return True
+        else:
+            return False
+
+
+def split_mongo_samples_by_version(
+    samples: List[Sample], cp_samples_df_v0: DataFrame, cp_samples_df_v1: DataFrame
+):  # noqa: E501
```
Chris-Friend, line 157: Can specify the return type here, and update description. I think the type should be `Dict[FilteredPositiveIdentifier, List[Sample]]`

migrations/helpers/update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Returns:
+        samples_by_version {Dict[List[Sample]]} -- Samples split by version
+    """
+    v0_cp_samples = cp_samples_df_v0[[FIELD_ROOT_SAMPLE_ID, FIELD_PLATE_BARCODE]].to_numpy().tolist()  # noqa: E501
+    v1_cp_samples = cp_samples_df_v1[[FIELD_ROOT_SAMPLE_ID, FIELD_PLATE_BARCODE]].to_numpy().tolist()  # noqa: E501
+
+    v0_unmigrated_samples = []
+    v1_unmigrated_samples = []
+    v2_unmigrated_samples = []
```
Chris-Friend, line 175: Could you rename these? As far as I understand the workflow can have migrated samples in (e.g. if it fails at a step after Mongo was updated with v0 samples); and this method shouldn't try to know about the migrated state of things - it should just split samples up

tests/migrations/helpers/test_update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,91 @@
...
+    split_mongo_samples_by_version,
+)
+from crawler.constants import (
+    FIELD_PLATE_BARCODE,
+    FIELD_ROOT_SAMPLE_ID,
+    V0_V1_CUTOFF_TIMESTAMP,
+    V1_V2_CUTOFF_TIMESTAMP,
+)
+
+# ----- migration helper function tests -----
```
Chris-Friend, line 16: Not urgent, but might be could to include some none happy path testing, e.g. when things throw, return empty arrays etc.

tests/conftest.py
```diff
@@ -144,6 +419,123 @@ def testing_samples(samples_collection_accessor):
...
+
+
+@pytest.fixture
+def event_wh_sql_engine(config):
+    create_engine_string = f"mysql+pymysql://{config.WAREHOUSES_RW_CONN_STRING}/{config.EVENTS_WH_DB}"
+    return sqlalchemy.create_engine(create_engine_string, pool_recycle=3600)
+
+
+@pytest.fixture
+def mlwh_sql_engine(config):
```
Chris-Friend, line 534: As mentioned, not sure if `event_wh_sql_engine` and `mlwh_sql_engine` need to be test fixtures. Could be wrong though

migrations/update_legacy_filtered_positives.py
```diff
@@ -0,0 +1,171 @@
...
+                        run before - do you still wish to proceed? (yes/no):"
+            continue_migration = get_input(question)
+
+            if continue_migration == "yes":
+                pass
+            elif continue_migration == "no":
+                logger.info("Now exiting migration")
+                raise Exception()
+            else:
+                logger.info("Invalid input, please enter 'yes' or 'no'. Now exiting migration")
```
Chris-Friend, line 76: I think this should be a warning

migrations/update_legacy_filtered_positives.py
```diff
@@ -0,0 +1,171 @@
...
+                raise Exception()
```
Chris-Friend, line 77: I don't particularly like that exceptions are raised here. Could you instead set a local bool variable that you check before line 79 that just skips lines 79-140, resulting in hitting the `finally` block without having to raise an exception?

migrations/update_legacy_filtered_positives.py
```diff
@@ -0,0 +1,171 @@
...
+from crawler.constants import (
+    V0_V1_CUTOFF_TIMESTAMP,
+    V1_V2_CUTOFF_TIMESTAMP,
+)
+from crawler.filtered_positive_identifier import (
+    FILTERED_POSITIVE_VERSION_0,
+    FILTERED_POSITIVE_VERSION_1,
+    FILTERED_POSITIVE_VERSION_2,
+)
+from migrations.helpers.dart_samples_update_helper import extract_required_cp_info
```
Chris-Friend, line 25: I move this method in my open PR to `shared_helper.py`. Just be aware that one of us may have to change this depending who's branch gets merged in first

tests/migrations/test_update_legacy_filtered_positives.py
```diff
@@ -0,0 +1,193 @@
...
+        mock_user_input.return_value = "no"
+        mock_update_mongo, mock_update_mlwh = mock_helper_database_updates
+
+        update_legacy_filtered_positives.run("crawler.config.integration")
+
+        mock_update_mongo.assert_not_called()
+        mock_update_mlwh.assert_not_called()
+
+
+def test_update_legacy_filtered_positives_exception_raised_if_user_enters_invalid_input(
```
Chris-Friend, line 83: As in the migration file, think this and the first test should just check no database is updated and not raise an exception to skip the code

tests/migrations/test_update_legacy_filtered_positives.py
```diff
@@ -0,0 +1,193 @@
...
+        FilteredPositiveIdentifierV2(): [{"plate_barcode": "3"}],
+    }
+    mock_update_mongo.return_value = True
+    mock_update_mlwh.return_value = True
+
+    update_legacy_filtered_positives.run("crawler.config.integration")
+
+    assert mock_update_filtered_positive_fields.call_count == 3
+    assert mock_update_mongo.call_count == 3
+    assert mock_update_mlwh.call_count == 3
```
Chris-Friend, line 160: Worth also checking what arguments these methods were called with, e.g. the correct list of samples?

Also in below test in which user input is required

tests/migrations/test_update_legacy_filtered_positives.py
```diff
@@ -0,0 +1,193 @@
...
+    mock_v0_version_set.return_value = False
+    mock_legacy_mongo_samples.return_value = [{"plate_barcode": "1"}]
+    mock_extract_required_cp_info.return_value = [["id_1"], ["plate_barcode_1"]]
+    mock_get_cherrypicked_samples_by_date.side_effect = Exception("Boom!")
+
+    with pytest.raises(Exception):
+        update_legacy_filtered_positives.run("crawler.config.integration")
+        mock_update_mongo.assert_not_called()
+        mock_update_mlwh.assert_not_called()
+
```
Chris-Friend, line 130: Might also be worth testing no database updates occur when:
* extracting cherrypicking info throws
* getting cherrypicked samples by date throws
* splitting the samples throws
* updating the filtered positive values throws

Chris-Friend (commented):
Some good work here. A complex migration, but easy to follow the logic used

crawler/config/test.py
```diff
@@ -56,5 +59,24 @@
 DART_DB_DRIVER = "{ODBC Driver 17 for SQL Server}"
 
 # logging config
+LOGGING["loggers"]["crawler"]["level"] = "DEBUG"  # noqa: F405
+LOGGING["loggers"]["crawler"]["handlers"] = ["colored_stream"]  # noqa: F405
```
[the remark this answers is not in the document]
alexperrin1: Yep you're right, happened when I merged develop in. Removed the duplicate

crawler/config/test.py
```diff
@@ -56,5 +59,24 @@
...
+
+# Event warehouse database details
+EVENT_WH_SUBJECTS_TABLE = "subjects"
+EVENT_WH_ROLES_TABLE = "roles"
+EVENT_WH_EVENTS_TABLE = "events"
+EVENT_WH_EVENT_TYPES_TABLE = "event_types"
+EVENT_WH_SUBJECT_TYPES_TABLE = "subject_types"
+EVENT_WH_ROLE_TYPES_TABLE = "role_types"
+
+EVENTS_WH_DB = "event_warehouse_test"
```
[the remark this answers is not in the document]
alexperrin1: I've added a comment to mention this

crawler/db.py
```diff
@@ -389,3 +390,8 @@ def add_dart_plate_if_doesnt_exist(cursor: pyodbc.Cursor, plate_barcode: str, bi
         raise DartStateError(f"DART plate {plate_barcode} should have a state")
 
     return state
+
+
+def create_mysql_connection_engine(connection_string: str, database: str) -> Engine:
+    create_engine_string = f"mysql+pymysql://{connection_string}/{database}"
+    return sqlalchemy.create_engine(create_engine_string, pool_recycle=3600)
```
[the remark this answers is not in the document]
alexperrin1: I've added a couple of tests to check that the right type is returned, and that the engine can connect

migrations/update_legacy_filtered_positives.py
alexperrin1: Yep, makes sense, I've changed to use a bool instead

migrations/update_legacy_filtered_positives.py
alexperrin1: OK 👍

tests/migrations/test_update_legacy_filtered_positives.py
alexperrin1: I've changed these tests

tests/migrations/test_update_legacy_filtered_positives.py
alexperrin1: 👍  good point, done this

tests/migrations/test_update_legacy_filtered_positives.py
alexperrin1: I've added three tests - already had the `get_cherrypicked_samples_by_date` tested, but renamed to make more clear

Chris-Friend:
There are still a few unresolved comments on the PR. You might need to "Load hidden conversations". Would you be able to look into these?
* method documentation changes
* test streamlining/robustness
* using `count_documents` instead of `find` so we aren't unnecessarily pulling all samples into memory
* behaviour when a sample sits exactly on the cut-off
* some other minor things

migrations/helpers/update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,191 @@
...
+
+def legacy_mongo_samples(config: ModuleType):
+    """Gets all samples from Mongo which have not had the filtered positive field set by Crawler
+
+    Arguments:
+        None
+
+    Returns:
+        List[Dict] -- All unmigrated samples from Mongo
+    """
```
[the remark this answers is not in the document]
alexperrin1: Updated 👍

tests/migrations/helpers/test_update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_legacy_mongo_samples_returns_correct_samples_filtered_by_date(
+    config, filtered_positive_testing_samples
+):  # noqa: E501
+    result = legacy_mongo_samples(config)
+    assert len(result) == 3
```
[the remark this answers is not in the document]
alexperrin1: I've put in a check for this

migrations/helpers/update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,191 @@
...
+) -> Optional[DataFrame]:
+    """Find which samples have been cherrypicked between defined dates using MLWH & Events warehouse.
+    Returns dataframe with 4 columns: those needed to uniquely identify the sample resulting
+    dataframe only contains those samples that have been cherrypicked (those that have an entry
+    for the relevant event type in the event warehouse)
+
+    Args:
+        root_sample_ids (List[str]): [description]
+        plate_barcodes (List[str]): [description]
+        chunk_size (int, optional): [description]. Defaults to 50000.
```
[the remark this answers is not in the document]
alexperrin1: Added these 👍

tests/migrations/helpers/test_update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_check_versions_set_returns_false_with_no_v0_samples(
+    config, filtered_positive_testing_samples_no_v0
+):  # noqa: E501
+    assert v0_version_set(config) is False
+
+
+def test_get_cherrypicked_samples_by_date_v0_returns_expected(
+    config, event_wh_data, mlwh_sample_stock_resource, mlwh_sql_engine, event_wh_sql_engine
```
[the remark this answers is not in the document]
alexperrin1: You're right about them not being needed here, I've removed them.

For the way that `event_wh_data` is set up in conftest.py, it assumes that the engine is a fixture (this is the same in Lighthouse). I **think** that you could change the way it's used in there so that it's just a function and not a fixture. Might then be useful to update in Lighthouse as well, so that they are consistent. Worth doing?

tests/migrations/helpers/test_update_legacy_filtered_positives_helper.py
```diff
@@ -0,0 +1,91 @@
...
+    pd.testing.assert_frame_equal(expected, returned_samples)
+
+
+def test_get_cherrypicked_samples_by_date_v1_returns_expected(
+    config, event_wh_data, mlwh_sample_stock_resource, mlwh_sql_engine, event_wh_sql_engine
+):
+    root_sample_ids = ["root_1", "root_2", "root_3", "root_4"]
+    plate_barcodes = ["pb_1", "pb_2", "pb_3", "pb_4"]
+
+    expected_rows = [["root_1", "pb_1"], ["root_3", "pb_3"]]
```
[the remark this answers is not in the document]
alexperrin1: Nice spot - I've updated the query so that it looks like this:

f" AND mlwh_sample.created >= '{start_date}'"
f" AND mlwh_sample.created < '{end_date}'"

Anything created before and any time up to but not including the cut-off is defined as the version in question - does that make sense?

migrations/helpers/update_legacy_filtered_positives_helper.py
alexperrin1: Updated

migrations/helpers/update_legacy_filtered_positives_helper.py
alexperrin1: Good idea - changed this

Chris-Friend: You can still just `return num_v0_samples > 0` rather than the if-statement

tests/migrations/helpers/test_update_legacy_filtered_positives_helper.py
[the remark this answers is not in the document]
Chris-Friend: Yep I think this makes sense. The upper bound is not inclusive, the lower bound is :+1:

migrations/helpers/update_legacy_filtered_positives_helper.py
alexperrin1: Oh OK, see what you mean, done this. Had to also change tests to use `==` rather than `is` in assert

migrations/helpers/update_legacy_filtered_positives_helper.py
alexperrin1: Renamed 👍

tests/migrations/helpers/test_update_legacy_filtered_positives_helper.py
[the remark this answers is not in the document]
Chris-Friend: Nah keep it as a test fixture. I don't think there is anything wrong with test fixture inheritance (of sorts)

tests/conftest.py
Chris-Friend: As discussed in a previous comment, keeping these as test fixtures

migrations/helpers/update_legacy_filtered_positives_helper.py
Chris-Friend: Although you forgot to include the `> 0` in the return statement. That was why you had to change the tests 🙂

andrewsparkes (approved):

PR #190, opened 2020-12-21
Update docker compose

stevieing:
Closes #

*
*
* ...


KatyTaylor:
From testing it locally, it is not using the MySQL service from the docker container, it is using my local one. 

When I did `brew services stop MySQL`, to remove the local MySQL, the crawler tests (from within the crawler container) failed with the following:
`ERROR    crawler.db:db.py:200 Exception on connecting to MySQL database: 1049 (42000): Unknown database 'unified_warehouse_test'`

Tried to setup the database by running `python ./setup_test_db.py`, but got similar error when running this script:
`Exception on connecting to MySQL database: 1049 (42000): Unknown database 'unified_warehouse_test'`

This seems to come from db.py, method `create_mysql_connection`, where it is trying to connect using the database name, before we have actually created the db. However, if this script is faulty, not sure why it is not failing in the ci builds.

crawler/constants.py
```diff
@@ -119,6 +120,12 @@
 # DART others
 DART_SET_PROP_STATUS_SUCCESS = 0
 
+# If we're running in a container, then instead of localhost
```
pjvv, line 123: I don't think we should be putting connection config in the constants file. You could just move `os.environ.get("LOCALHOST", "127.0.0.1")` to the default config file. Test imports all config from defaults.

docker-compose.yml
```diff
@@ -32,6 +32,17 @@ services:
       ACCEPT_EULA: Y
       SA_PASSWORD: MyS3cr3tPassw0rd
 
+  # Setup the runner
+  # we need to set the localhost and root password as they can be different
+  runner:
```
pjvv, line 37: I've found it clearer to name the service after the app, i.e. crawler.

pjvv (changes requested):

docker-compose.yml
KatyTaylor: Hi Phil, they all seem to be prefixed by crawer, see screenshot:
Screenshot 2021-01-05 at 14 37 30

So if changed to 'crawler' you get 'crawler_crawler'

pjvv: Yip, that's true, root name is used to prepend all the services in the "stack" but imagine you add another service (in a larger stack) to the stack which behaves in a similar manner, calling it by it's name instead of how/what it does is clearer. I think I'm arguing from a consistence point of view because that is how other stacks are defined.

KatyTaylor: OK no probs

crawler/constants.py
KatyTaylor: What's your opinion on the ROOT_PASSWORD one, @pjvv ? same deal?

pjvv: Oh yeah, same, sorry was just being lazy and used that one as an example!

stevieing:
> > From testing it locally, it is not using the MySQL service from the docker container, it is using my local one.
> > When I did `brew services stop mysql`, to remove the local MySQL, the crawler tests (from within the crawler container) failed with the following:
> > `ERROR crawler.db:db.py:200 Exception on connecting to MySQL database: 1049 (42000): Unknown database 'unified_warehouse_test'`
> > Tried to setup the database by running `python ./setup_test_db.py`, but got similar error when running this script:
> > `Exception on connecting to MySQL database: 1049 (42000): Unknown database 'unified_warehouse_test'`
> > This seems to come from db.py, method `create_mysql_connection`, where it is trying to connect using the database name, before we have actually created the db. However, if this script is faulty, not sure why it is not failing in the ci builds.
> 
> I will try this out to see what happens on mine.
> 
> I'm thinking it will be worth getting a feel for who wants to use Docker for development and who just for the ancillary services. I've used it both ways and find it quicker and easier to develop locally with pipenv and then spin up the services with Docker. This compose file will build everything but maybe most of us just want the services?

I can put this succintly. I can't get it to work without docker so I have no choice. I spent 3 days before Xmas to get crawler up and running using pipenv and there was no solution. From my point of view if I can't use docker then I can't work on crawler.

pjvv:
Where were you two years ago?! Would have loved this enthusiasm for docker!

KatyTaylor:
> > From testing it locally, it is not using the MySQL service from the docker container, it is using my local one.
> > When I did `brew services stop mysql`, to remove the local MySQL, the crawler tests (from within the crawler container) failed with the following:
> > `ERROR crawler.db:db.py:200 Exception on connecting to MySQL database: 1049 (42000): Unknown database 'unified_warehouse_test'`
> > Tried to setup the database by running `python ./setup_test_db.py`, but got similar error when running this script:
> > `Exception on connecting to MySQL database: 1049 (42000): Unknown database 'unified_warehouse_test'`
> > This seems to come from db.py, method `create_mysql_connection`, where it is trying to connect using the database name, before we have actually created the db. However, if this script is faulty, not sure why it is not failing in the ci builds.
> 
> I will try this out to see what happens on mine.
> 
> I'm thinking it will be worth getting a feel for who wants to use Docker for development and who just for the ancillary services. I've used it both ways and find it quicker and easier to develop locally with pipenv and then spin up the services with Docker. This compose file will build everything but maybe most of us just want the services?

I wrote this longer reply before Steve put it succinctly 😆  will leave it here in case helpful...

Thanks! Yes, I'd be up for having a discussion with the team about who's using what.

I can't currently run it locally because I get an error while importing` pyodbc`, which I think at least Steve and Eduardo have also got - that's also an issue with Lighthouse, and I think that's why we started using docker locally for this. Interested to know how you got round that? 

When I run the tests in the container using the method Steve's written in the Readme, it is using my local MySQL and also MongoDB - so it currently only makes sense to put the MS SQL in the docker compose.

README.md
```diff
@@ -215,6 +215,23 @@ A little convenience script can be used to run the formatting, type checking and
...
+
+    docker-compose up
+
+To run the tests:
+
+You will need to find the id of the container with image name crawler_runner
+
+    docker exec -ti <container_id> python -m pytest -vs
+
+There is now a volume for the runner so there is hot reloading i.e. changes in the code and tests will be updated when you rerun tests.
```
KatyTaylor, line 233: @stevieing , how does the hot reloading work? I tried changing some test code locally, saved and re-ran the tests in the container, and it didn't reflect the changes - do I need to change the code elsewhere?

Thanks

stevieing: That's weird. It should just work. Without the volume it didn't work. I will try it just to check whether it still works for me.

KatyTaylor: I take it back - it works! 🤦

pjvv (approved):

PR #193, opened 2020-12-24
Gpl 770 beckman cherrypicked samples

Chris-Friend:
* Add to the `get_cherrypicked_samples` method a query to identify samples that have been cherrypicked by the Beckman workflow.
* Also lots of tesst to check the complexities


andrewsparkes (commented):
Similar comments to the other PR, but looks fine!

PR #197, opened 2021-01-06
Minor refactor

pjvv:

* add more typing hints


crawler/centre_file_state.py
```diff
@@ -0,0 +1,11 @@
+from enum import Enum, auto
```
KatyTaylor, line 1: Nice, I wanted these to be separate files 👍

pjvv: I wanted to get the Centre and CentreFile in separate files too because the `file_processing.py` file is getting out of hand but have deferred that to a larger refactor.

crawler/db.py
```diff
@@ -189,16 +191,12 @@ def create_mysql_connection(config: ModuleType, readonly=True) -> CMySQLConnecti
...
-                logger.error("MySQL Connection Failed")
+        logger.debug("MySQL Connection Successful")
 
+        return cast(CMySQLConnection, mysql_conn)
     except mysql.Error as e:
-        logger.error(f"Exception on connecting to MySQL database: {e}")
-
-    return mysql_conn
+        logger.error(f"Exception while connecting to MySQL database: {e}")
+        raise
```
KatyTaylor, line 199: Not sure you should change the behaviour here without changing the error handling of the code that calls it - have you followed it up the chain?

crawler/file_processing.py
```diff
@@ -1138,48 +1157,48 @@ def row_channel_result_values_valid(self, row, line_number) -> bool:
...
-            num {Decimal128} - the number to be tested
+            num {ModifiedRowValue} - the number to be tested
 
         Returns:
             bool - whether the value lies within range
         """
-        # cannot compare Decimal128 to Decimal or to other Decimal128s
-        return range_min <= num.to_decimal() <= range_max
+        # TODO: this casting is dangerous and needs to be re-thought...
+        return range_min <= Decimal(cast(Decimal128, num).to_decimal()) <= range_max
```
KatyTaylor, line 1173: retain the other comment too?

KatyTaylor: So this `cast` function is purely for the type checker? 
Is that also the reason why you've wrapped it all in `Decimal()` ?

Seems a shame to make the code less readable for the type checker. Might as well use a strongly typed language? I haven't read about it, explain it to me :)

crawler/file_processing.py
```diff
@@ -1138,48 +1157,48 @@ def row_channel_result_values_valid(self, row, line_number) -> bool:
...
         """
-        if row.get(fieldname):
-            if not self.is_within_cq_range(MIN_CQ_VALUE, MAX_CQ_VALUE, row.get(fieldname)):
-                self.logging_collection.add_error(
-                    "TYPE 20",
-                    f"{fieldname} not in range ({MIN_CQ_VALUE}, {MAX_CQ_VALUE}), "
-                    f"line: {line_number}, result: {row.get(fieldname)}",
-                )
-                return False
+        if fieldname in row and not self.is_within_cq_range(MIN_CQ_VALUE, MAX_CQ_VALUE, row[fieldname]):
```
KatyTaylor, line 1186: This pattern was changed previously - see this commit - https://github.com/sanger/crawler/commit/fcc23fc07a6448d9cf2d7206966044197c2832cf 

I am trying to remember what the difference is but I think we thought quite hard about it before.

Maybe `row.get(fieldname)` checks if `fieldname in row` and `row[fieldname] is not None` or something?

If it turns out that `row.get(fieldname)` turns out to be correct, probably worth putting in some tests to avoid this back and forth in future?

crawler/centre_file_state.py
KatyTaylor: yep sounds good

crawler/file_processing.py
pjvv: Yes thanks, I will add the previous helpful comment back.

I'm no expert, I'm on a journey with mypy and typing in python - it's quite new, at least to me. :P

Yes it's a shame that in this instance the readability is affected, but in general, the type annotations are helpful while coding and for catching obvious bugs. I prefer the dynamic typing of python (and python as a language) more than static typed languages like Java etc. and python's type annotations allows the benefit of types with a "scripting" language.

I have some ideas to refactor which should clean this up nicely.

crawler/db.py
pjvv: Good spot thanks, will dig a little deeper.

crawler/file_processing.py
pjvv: Oh OK, thanks and good remembering! Will have a look a little deeper here too and add comments to prevent back and forth too.

tests/test_file_processing.py
```diff
@@ -1010,48 +1006,6 @@ def test_file_name_date_parses_right(config):
...
-        assert len(errors_folder.listdir()) == 1
-
-        # check the file state again now the error version exists
-        centre_file.set_state_for_file()
-
-        assert centre_file.file_state == CentreFileState.FILE_PROCESSED_WITH_ERROR
-
-
-def test_set_state_for_file_when_in_success_folder(config):
-    return False
```
KatyTaylor, line 1052: Huh?!

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
KatyTaylor, line 12: Looks like you need a special plugin for your text editor for this to work? Could add instructions to the readme explaining how?

.env.example
```diff
@@ -1,11 +1 @@
-CENTRE_DETAILS_FILE_PATH=crawler/config/centre_details.json
-MONGO_DB=crawlerDevelopmentDB
-MONGO_HOST=127.0.0.1
-MONGO_PORT=27017
-SFTP_HOST=<sftp_host>
-SFTP_PASSWORD=<sftp_password>
```
KatyTaylor, line 6: was looking into these, and it seems SFTP_PASSWORD is obsolete, but mentioned in test.py - so maybe remove it from there too?

.env.example
```diff
@@ -1,11 +1 @@
...
-SFTP_PORT=22
-SFTP_USER=<sftp_user>
-SLACK_API_TOKEN=xoxb-123
-SLACK_CHANNEL_ID=Cxxx
-SETTINGS_MODULE=crawler.config.development
\ No newline at end of file
+SETTINGS_MODULE=crawler.config.development
```
KatyTaylor, line 1: Could be worth adding LOCALHOST=host.docker.internal in here, as that is needed if you run this inside docker?

Pipfile
```diff
@@ -13,6 +13,7 @@ pytest = "*"
 pytest-cov = "*"
 pytest-freezegun = "*"
 black = "*"
+monkeytype = "*"
```
KatyTaylor, line 16: Is this the sort of thing we want to be running manually regularly? If so, maybe add something to the readme about it?

crawler/config/defaults.py
```diff
@@ -105,6 +34,8 @@
 MLWH_DB_RW_USER = "root"
 MLWH_DB_RW_PASSWORD = ROOT_PASSWORD
 
+EVENTS_WH_DB = ""
```
KatyTaylor, line 37: why blank?

crawler/db.py
```diff
@@ -286,19 +284,21 @@ def create_dart_sql_server_conn(config: ModuleType) -> Optional[pyodbc.Connectio
 
     logger.debug(f"Attempting to connect to {dart_db_host} on port {dart_db_port}")
 
-    sql_server_conn = None
     try:
         sql_server_conn = pyodbc.connect(connection_string)
 
-        if sql_server_conn is not None:
+        if sql_server_conn:
```
KatyTaylor, line 290: Are we sure this is the same thing? Just aware that recently we had that thing about dataframes and truthiness... so are we sure that sql_server_conn falsiness is the same as not None?

migrations/helpers/shared_helper.py
```diff
@@ -81,16 +81,13 @@ def get_cherrypicked_samples(
         ]
 
         sql_engine = sqlalchemy.create_engine(
-            (
-                f"mysql+pymysql://{config.MLWH_DB_RO_USER}:{config.MLWH_DB_RO_PASSWORD}"  # type: ignore
-                f"@{config.MLWH_DB_HOST}"  # type: ignore
-            ),
+            (f"mysql+pymysql://{config.MLWH_DB_RO_USER}:{config.MLWH_DB_RO_PASSWORD}" f"@{config.MLWH_DB_HOST}"),
```
KatyTaylor, line 84: is this the bit we added the port to when doing the migration? Might need careful merging

crawler/file_processing.py
```diff
@@ -308,56 +297,63 @@ def checksum_match(self, dir_path: str) -> bool:
...
-                            "same checksum but different filename"
+                            f"Found an identical file {backup_filename} in path {dir_path} which has the same checksum "
+                            "but a different filename"
                         )
                     return True
         return False
 
-    def get_centre_from_db(self):
+    def get_centre_from_db(self) -> Dict[str, Any]:
+        """Gets a document from the mongo centre collection which desctibes a lighthouse centre.
```
KatyTaylor, line 317:
```suggestion
        """Gets a document from the mongo centre collection which describes a lighthouse centre.
```

crawler/file_processing.py
```diff
@@ -597,7 +603,7 @@ def insert_samples_from_docs_into_mongo_db(self, docs_to_insert: List[Dict[str,
             logger.exception(e)
             return []
 
-    def insert_samples_from_docs_into_mlwh(self, docs_to_insert) -> bool:
+    def insert_samples_from_docs_into_mlwh(self, docs_to_insert):
```
KatyTaylor, line 606: why remove the return type annotation?

.editorconfig
pjvv: Yeah it does require an IDE extension to take effect. What do you think? Should we leave it our the repo for now and leave it up to devs if they want to add one? We could then ignore it in the .gitignore. I've added it to the standards doc in https://github.com/sanger/.github/pull/2/files so we could get some more input.

.env.example
pjvv: Thanks, have removed it from test.py too.

.env.example
pjvv: But are you entering the virtual env (`pipenv shell`) in the containter?

Pipfile
pjvv: Yeah this is a tool I'm trying which creates type annotations automatically - I'll add something in the README.

crawler/config/defaults.py
pjvv: Added `event_warehouse_development`.

crawler/db.py
pjvv: Reverted until larger refactor.

crawler/file_processing.py
pjvv: This was part of my MonkeyType testing, to see if it would pick this up.

pjvv: Still in progres...

crawler/file_processing.py
pjvv: So it was my initial misunderstanding of the `.get()` method. I think it is more readable this way (checking for the key in the dict first) but using the `.get()` method is more readable on line 1082 - soon to be updated and pushed. What do you think?

migrations/helpers/shared_helper.py
pjvv: Indeed, will add the port here to make doubly sure too.

tests/test_file_processing.py
pjvv: Yip, somehow snuck in there!

pjvv: Maybe needed a TODO.

crawler/db.py
pjvv: Reverted to larger refactor and from quick inspection does not seem like the correct behaviour to raise here.

emrojo (commented):
Some more coments

crawler/file_processing.py
pjvv: Yes, `row.get(fieldname)` returns the value of the dict with key fieldname (or None if there is no such key) but that value could be a boolean which is misleading. I think it's more readable to first check if that key exists and then do the necessary checks on the logic?

.editorconfig
pjvv: Will leave it in until dissent.

crawler/file_processing.py
pjvv: Can leave it for now as a "gardening" task along with another story.

crawler/file_processing.py
```diff
@@ -920,27 +945,35 @@ def parse_and_format_file_rows(self, csvreader: DictReader) -> Any:
 
             line_number += 1
 
-        logger.info(f"Rows with invalid structure in this file = {invalid_rows_count}")
-        logger.info(f"Rows that failed validation in this file = {failed_validation_count}")
+        logger.log(
+            INFO if invalid_rows_count == 0 else WARN,
+            f"Rows with invalid structure in this file = {invalid_rows_count}",
+        )
```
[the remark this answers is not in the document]
pjvv: Not sure what you mean? Would you rather have it at INFO?

crawler/file_processing.py
```diff
@@ -920,27 +945,35 @@ def parse_and_format_file_rows(self, csvreader: DictReader) -> Any:
...
+        logger.log(
+            INFO if failed_validation_count == 0 else WARN,
+            f"Rows that failed validation in this file = {failed_validation_count}",
```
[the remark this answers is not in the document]
pjvv: 👍🏼

crawler/helpers/general_helpers.py
```diff
@@ -254,7 +261,7 @@ def parse_decimal128(value: Decimal128) -> Optional[Decimal]:
         Decimal -- converted number
     """
     try:
-        return value.to_decimal()
+        return Decimal(value.to_decimal())
```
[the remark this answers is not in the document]
pjvv: The cast function is just for typing, it doesn't do anything with the argument, simply passes it through but type checkers know what type the argument then is.

crawler/file_processing.py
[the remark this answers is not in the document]
emrojo: If a warn means receiving a message in slack everytime we are in this situation, it could be too verbose

crawler/helpers/general_helpers.py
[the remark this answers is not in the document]
emrojo: is value.to_decimal() not a Decimal already? If this is an error by the syntax checker, why is casted with Decimal instead of annotating the value.to_decimal as a Decimal for the syntax checker with cast()

crawler/file_processing.py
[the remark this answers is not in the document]
pjvv: Oh no, currently we are only seeing ERROR in slack.

crawler/helpers/general_helpers.py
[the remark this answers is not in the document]
pjvv: Sorry! I did not read your first question properly, I have used `cast()` instead, that makes more sense.

emrojo (approved):

PR #216, opened 2021-01-15
Gpl 771 test fixes

### LINKED ISSUE #167: GPL-771 Migrate all samples to use the new filtered_positive fields

alexperrin1:
Closes #167 

* Batch legacy migration by dates (1 week took approx 15 mins)
* Batch updates to Mongo in terms of total samples. Updates were already batched in filtered positive False/true
* Batch MLWH updates by filtered positives true/false


crawler/db.py
```diff
@@ -259,6 +261,72 @@ def run_mysql_executemany_query(mysql_conn: CMySQLConnection, sql_query: str, va
...
+        query_args {List[Any]} -- additional sql query arguments
+    """
+    # fetch the cursor from the DB connection
+    cursor = mysql_conn.cursor()
+
+    try:
+        # executing the query with values
+        num_formatting_args = len(formatting_args)
+
+        # BN. If ROWS_PER_QUERY value is too high, you may get '2006 (HY000): MySQL server has
```
Chris-Friend, line 282: Worth changing this comment to match local variable names?

crawler/db.py
```diff
@@ -259,6 +261,72 @@ def run_mysql_executemany_query(mysql_conn: CMySQLConnection, sql_query: str, va
         mysql_conn.close()
 
 
+def run_mysql_execute_formatted_query(
```
Chris-Friend, line 264: Would you be able to add some tests for this, if you haven't already?

crawler/sql_queries.py
```diff
@@ -83,8 +83,22 @@
 filtered_positive_version = %(filtered_positive_version)s,
 filtered_positive_timestamp = %(filtered_positive_timestamp)s
 WHERE mongodb_id = %(mongodb_id)s
+AND root_sample_id = %(root_sample_id)s
+AND rna_id = %(rna_id)s
+AND result = %(result)s
```
Chris-Friend, line 88: Why was it required to add these identifying conditions here? Is `mongodb_id` not indexed and unique?

migrations/helpers/update_filtered_positives_helper.py
```diff
@@ -161,36 +164,50 @@ def update_mongo_filtered_positive_fields(
...
-                }
-            },
-        )
-    return True
+
+        num_samples = len(samples)
+        SAMPLES_PER_QUERY = 15000
+        samples_index = 0
+        logger.debug(f"Attempting to update {num_samples} rows in Mongo in batches of {SAMPLES_PER_QUERY}")
+        while samples_index < num_samples:
```
Chris-Friend, line 173: Would it be possible to add tests to check the batching of this? If existing tests pass that is a good sign, but they definitely don't input 15,000+ samples into this method 🙂

tests/migrations/helpers/test_update_legacy_filtered_positives_helper.py
```diff
@@ -153,3 +165,81 @@ def test_split_mongo_samples_by_version(unmigrated_mongo_testing_samples):
...
+        "migrations.helpers.update_legacy_filtered_positives_helper.run_mysql_execute_formatted_query",
+        side_effect=Exception("Boom!"),
+    ):
+        with pytest.raises(Exception):
+            update_mlwh_filtered_positive_fields_batched(
+                config, migrated_mongo_testing_samples, version, update_timestamp
+            )
+
+
+def test_update_mlwh_filtered_positive_fields_batched_calls_to_update_samples(
```
Chris-Friend, line 199: This should be in the `db.py` as you are testing the result of the method you call in the `update_mlwh_filtered_positive_fields_batched` method.

I think here you should test that the `run_mysql_execute_formatted_query` is called with the expected parameters (ids, sql_arguments etc.)

migrations/helpers/update_legacy_filtered_positives_helper.py
```diff
@@ -199,3 +232,76 @@ def split_mongo_samples_by_version(
     }
 
     return samples_by_version
+
+
+def update_mlwh_filtered_positive_fields_batched(
```
Chris-Friend, line 237: As mentioned, I think the existing filtered positive rule change migration should call this more performant MLWH update method. Do you reckon it worth doing this as part of these changes, or a new "refactor" task, e.g. https://github.com/sanger/crawler/issues/173

tests/conftest.py
```diff
@@ -239,6 +251,54 @@ def testing_files_for_process(cleanup_backups):
     },
 ]
 
+MIGRATED_MONGO_TESTING_SAMPLES = [
```
Chris-Friend, line 254: I think this name is too vague: there are lots of migrations, perhaps `MONGO_TESTING_SAMPLES_WITH_FILTERED_POSITIVE_FIELDS`?

tests/conftest.py
```diff
@@ -239,6 +251,54 @@ def testing_files_for_process(cleanup_backups):
...
+        FIELD_ROOT_SAMPLE_ID: "MCM002",
+        FIELD_RNA_ID: "BBB123",
+        FIELD_FILTERED_POSITIVE: False,
+        FIELD_FILTERED_POSITIVE_VERSION: "v2",
+        FIELD_FILTERED_POSITIVE_TIMESTAMP: "2020-01-01T00:00:00.000Z",
+    },
+]
+
+
+MLWH_SAMPLES = [
```
Chris-Friend, line 278: Similarly here (vague naming), if these samples are specifically added as they have filtered positive fields

tests/conftest.py
```diff
@@ -639,6 +699,16 @@ def unmigrated_mongo_testing_samples(config, event_wh_sql_engine):
     return UNMIGRATED_MONGO_TESTING_SAMPLES
 
 
+@pytest.fixture
+def migrated_mongo_testing_samples(config, event_wh_sql_engine):
+    return MIGRATED_MONGO_TESTING_SAMPLES
+
+
+@pytest.fixture
+def mlwh_samples(config, event_wh_sql_engine):
```
Chris-Friend, line 708: These methods should also be renamed if renaming the test data

tests/migrations/test_update_legacy_filtered_positives.py
```diff
@@ -107,92 +116,168 @@ def test_update_legacy_filtered_positives_exception_raised_if_user_enters_invali
...
-            side_effect=Exception("Boom!"),
-        ):
-            mock_v0_version_set.return_value = False
-            mock_update_mongo, mock_update_mlwh = mock_helper_database_updates
-            update_legacy_filtered_positives.run("crawler.config.integration")
+        update_legacy_filtered_positives.run("crawler.config.integration", start_date_input, end_date_input)
 
-            mock_update_mongo.assert_not_called()
-            mock_update_mlwh.assert_not_called()
+    mock_update_mongo.assert_called_once()
```
Chris-Friend, line 190: I'm not entirely sure anything gets called after an exception is caught in the `with pytest.raises(Exception):` block. Would you be able to check this (by ensuring the test fails if you assert a call was made)?

Chris-Friend (commented):
Looks good. No major comments

migrations/helpers/update_legacy_filtered_positives_helper.py
Chris-Friend: Actually I think @KatyTaylor was creating a new task with filtered positive rule change migration changes, so I think that can be done as part of that task. I'm not sure if that task has been created yet though

crawler/sql_queries.py
alexperrin1: This was incorrectly committed -  we initially added these to see if it would speed up the query. Now removed

crawler/db.py
alexperrin1: Yep done 👍

crawler/db.py
alexperrin1: I've added two tests, similar to those for the executemany query

migrations/helpers/update_filtered_positives_helper.py
Chris-Friend: This has been development-testing, and will be tested again during UAT testing

tests/migrations/test_update_legacy_filtered_positives.py
Chris-Friend: Looks like I'm wrong on this one: inverting the check causes the tests to fail, indicating that the assertions are called

Chris-Friend:
@KatyTaylor @pjvv I've made the changes Alex wasn't able to finish. Would you mind looking over the PR to check everything is in order?

migrations/helpers/update_legacy_filtered_positives_helper.py
pjvv: @KatyTaylor, is this still to be done?

pjvv (approved):

PR #226, opened 2021-01-20
GPL-826 second crawler run

KatyTaylor:


emrojo (approved):

runner.py
```diff
@@ -47,11 +47,13 @@
     if args.once:
         main.run(args.sftp, args.keep_files, args.add_to_dart)
     else:
-        time_to_run = "01:00"
-        print(f"Scheduled to run at {time_to_run}")
-        schedule.every().day.at(time_to_run).do(
-            main.run, sftp=args.sftp, keep_files=args.keep_files, add_to_dart=args.add_to_dart
-        )
+        times_to_run = ["01:00", "07:30"]
+        print(f"Scheduled to run at {times_to_run}")
```
harrietc52, line 51: `times_to_run.join(', ')` maybe?

harrietc52 (approved):
👍 minor comment

runner.py
KatyTaylor: I've left it as wanted to get it out, and it appears like the following, which is OK for logging I think:

Scheduled to run at ['01:00', '07:30']

PR #240, opened 2021-01-29
GPL-697 new lh lab randox

### LINKED ISSUE #138: GPL-697 New LH lab sample metadata & CT values (Randox) new files uploaded on 26th
**User story**
GPL-697 | As Head of Service Delivery for Genomic Surveillance (Sonia G) I would like for Randox lab to be able to submit sample metadata including CT values to Sanger to onboard a new lighthouse lab

Keith A (PAM)
Sonia G

**Acceptance criteria**
Basic new lab onboarding:
- [x] SFTP folder set up share name 'project-heron_randox'
- [ ] crawler processes files from that new folder
- [ ] sequencescape accepts plates with the correct barcode format ('URNxxxxx')
- [ ] baracoda creates COG UK IDs with the correct prefix ('RAND')

Adaptations to Randox files (TBD - currently asking if they are able to adapt their files to us):
- [ ] translate their column names into our standard column names
- [ ] translate Plate Well Number integer value into well position e.g. 'A01'
- [ ] create RNA ID value by concatenating Batch Name with above well position
- [ ] Make sure upper / lower case in Result field doesn't matter
- [ ] Check Date Tested format is consistent with others
- [ ] Allow 'Undetermined' value in CT result fields
- [ ] Allow 'E gene' value in CT target fields

KatyTaylor:
Closes #138 

* Add centre config for Randox lab
* Add 'E-Gene' as an allowed CT target value
* Run through the headers of the file first, and correct any that are nearly right (Root Sample -> Root Sample ID)
* Add method to remove byte order mark (BOM) if present
* Made Viral Prep ID and RNA-PCR ID not required


crawler/file_processing.py
```diff
@@ -750,6 +780,23 @@ def get_channel_headers_mapping(self) -> Dict[str, str]:
...
+        if csvreader.fieldnames:
+            for i in range(0, len(csvreader.fieldnames)):
+                for reg in self.header_regex_correction_dict.keys():
+                    match = re.match(reg, csvreader.fieldnames[i])
+                    if match:
+                        logger.warning(
+                            f"Found '{reg}' in field name '{csvreader.fieldnames[i]}', "
+                            f"correcting to '{self.header_regex_correction_dict[reg]}'"
+                        )
+                        csvreader.fieldnames[i] = self.header_regex_correction_dict[reg]
```
KatyTaylor, line 798: mypy fails here with error:
`Unsupported target for indexed assignment ("Sequence[str]")`

I think this is because the type 'DictReader' is not expecting to have its items assigned (read-only).

It does feel a bit wrong to be modifying this object, what do others think?

I did it this way - a quick run through to 'correct' any headers that are a bit wrong - because that means none of the validating or parsing code needs to change - they can just assume the headers are correct.

Any thoughts welcome, thanks.

KatyTaylor:
Think this will also need a modification in Sequencescape, for the new foreign barcode format

tests/test_file_processing.py
```diff
@@ -913,6 +913,36 @@ def test_where_positive_result_does_not_align_with_ct_channel_results(config):
...
+        # construct a bytes object containing a byte order mark (BOM)
+        header_with_bom = b"\xef\xbb\xbfRoot Sample ID"
+        bom_as_utf8_string = header_with_bom.decode("utf-8")
+
+        fake_csv.write(f"{bom_as_utf8_string},RNA ID\n")
+        fake_csv.write("1,RNA_0043_\n")
+        fake_csv.seek(0)
+
+        csv_to_test_reader = DictReader(fake_csv)
+        assert csv_to_test_reader.fieldnames != ["Root Sample ID", "RNA ID"]
```
pjvv, line 927: I think you're just checking the standard library here.

crawler/file_processing.py
```diff
@@ -255,13 +265,15 @@ def __init__(self, file_name: str, centre: Centre):
...
-            FIELD_VIRAL_PREP_ID,
             FIELD_RNA_ID,
-            FIELD_RNA_PCR_ID,
             FIELD_RESULT,
             FIELD_DATE_TESTED,
         }
 
+        # These are to allow some variability in headers,
+        # due to receiving inconsistent file formats
+        self.header_regex_correction_dict = {"Root Sample": FIELD_ROOT_SAMPLE_ID}
```
pjvv, line 275: Should we being this lenient? Shouldn't it only match "Root Sample" or "Root Sample ID"? Something like: `"^Root Sample([ ]?ID) $"`

pjvv (changes requested):

crawler/file_processing.py
```diff
@@ -750,6 +780,23 @@ def get_channel_headers_mapping(self) -> Dict[str, str]:
...
         return self.CHANNEL_FIELDS_MAPPING
 
+    def correct_headers(self, csvreader: DictReader) -> None:
+        """Checks for any headers in the CSV file that are wrong but recognisable, and fixes them.
+        Necessary due to variability in the file format we receive.
+        """
+        logger.debug("Checking CSV for nearly-correct header names and fixing them")
+
+        if csvreader.fieldnames:
+            for i in range(0, len(csvreader.fieldnames)):
```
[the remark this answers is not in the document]
KatyTaylor: Ah, will check it out, thanks

crawler/file_processing.py
KatyTaylor: I'm not sure really. I'd err on the side of leaving it more lenient, just because we've received several iterations of the file already, so I wouldn't be surprised if one turns up in future with 'Root Sample Number' or something :)

pjvv: 🤔 True, but at the same time, is it possible that they send something unrelated to the "ID" but related to the "Root Sample"? Like "Root Sample Result" or "Root Sample Date Tested"? I hesitant to allow a bug on their side to pollute data on our side but it could be very edge case...

tests/test_file_processing.py
KatyTaylor: I'm doing this check here to make sure that we initially don't get these column headers, so we know that when we reach the second assertion on line 930, we know for sure that the remove_bom method is responsible for the change to the correct column headers.

On top of that, I found checking for the presence of a BOM a bit slippery! Some standard python functions seemed to auto-remove it, so it looked like it wasn't there when it was (specifically, when they were in a list, you could see the BOM, but when you selected an item from the list via indexing, you could no longer see the BOM). So it's good to have an assertion in there proving the BOM is present, before we prove it's not present.

Similar answer to line 940 (minus the BOM bit). 

Does that make sense or have I misunderstood your comment?

crawler/file_processing.py
pjvv: 🤔 

- what about assigning another attribute which we can check later? `corrected_fieldnames`?
- Another option, subclassing DictReader. 
- Another, adding a boolean attribute `is_fieldnames_corrected` then check `corrected_fieldnames` when needed else just `fieldnames`.

tests/test_file_processing.py
```diff
@@ -913,6 +913,36 @@ def test_where_positive_result_does_not_align_with_ct_channel_results(config):
...
+
+
+def test_correct_headers(centre_file):
+    with StringIO() as fake_csv:
+        fake_csv.write("Root Sample Wrong Name,RNA ID\n")
+        fake_csv.write("1,RNA_0043_\n")
+        fake_csv.seek(0)
+
+        csv_to_test_reader = DictReader(fake_csv)
+        assert csv_to_test_reader.fieldnames == ["Root Sample Wrong Name", "RNA ID"]
```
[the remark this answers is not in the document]
pjvv: This will be a good test if you subclass DictReader!

crawler/file_processing.py
KatyTaylor: Yep OK, that is possible. I'll try to modify it so it allows 'Root Sample' and 'Root Sample ID', and with trailing whitespace or something

tests/test_file_processing.py
pjvv: OK, yes thank you, that does make sense!

crawler/file_processing.py
KatyTaylor: OK so I've been thinking about this one some more. I misunderstood the error in my previous message. 

It actually means that it thinks `csvreader.fieldnames` is of the type `Sequence[str]`, and for some types of sequence you can't assign values to items in the sequence using indexing (e.g. you can't do `myString[1] = 'f'` or myTuple[1] = 'thing', but you can do `myList[1]` = 'thing')

However, it is actually of the type `list`, so indexed assignment is fine.

So now I think this is mypy's problem, and I've added a # type: ignore to it.

What do you think?

crawler/file_processing.py
```diff
@@ -272,7 +272,7 @@ def __init__(self, file_name: str, centre: Centre):
 
         # These are to allow some variability in headers,
         # due to receiving inconsistent file formats
-        self.header_regex_correction_dict = {"Root Sample": FIELD_ROOT_SAMPLE_ID}
+        self.header_regex_correction_dict = {r"^\s*Root Sample\s*$": FIELD_ROOT_SAMPLE_ID}
```
pjvv, line 275: I think we should strip the fieldnames first before looking for matches to make our regex simpler. This would require a little rework in the crawler though so you could just do it for this change to start?

crawler/file_processing.py
```diff
@@ -758,7 +758,7 @@ def remove_bom(self, csvreader: DictReader) -> None:
 
             if has_bom:
```
pjvv, line 759: This could just be `if as_bytes_from_utf8[:3] == b"\xef\xbb\xbf":`?

pjvv (dismissed):

crawler/file_processing.py
KatyTaylor: I think in this case I like declaring the variable has_bom, because it's unclear what `as_bytes_from_utf8[:3] == b"\xef\xbb\xbf"` actually means otherwise

crawler/file_processing.py
KatyTaylor: Yes that would be nicer, will take a look at what I can change here

pjvv (approved):

PR #257, opened 2021-02-10
Gpl854 cambridge rna id padding

### LINKED ISSUE #253: GPL-854 parse non zero padded co-ordinates in 'RNA ID' values in plate maps from Biomeks (Cambridge)
**User story**
GPL-854 | As LH Stakeholder (Alan K) I would like the well ordinate on 'RNA ID' to parse non zero padded co-ordinates because is a Biomek issue and cannot be fixed in-house at Cambridge (and Biomek consolidate plate are future strategy for LH)

Alan K (PAM)
Rich L

**Context**
Cambridge moved from plates of positives, negatives and voids to consolidated plates on 4th February. Where previously the plate map RNA ID had plate map and well position concatenated with 0 padding eg EGC000035_A01 this is no longer the case with Biomek. RNA ID with no padding, now needs to be parsed eg EGC000035_A1

The is resulting in all samples being rejected as type 9 errors:

- http://lighthouse-ui.psd.sanger.ac.uk/imports (search on CB, look dates from 4th)
- Failed regex on plate barcode errors (TYPE 9): https://ssg-confluence.internal.sanger.ac.uk/x/jqsKBg (table 3.1)

Alan K wrote (Tuesday, 9 February 2021 at 09:39) 

_Padded zero issue is a Biomek issue and cannot be fixed in-house at Cambridge.
 
Because of this they supposedly checked directly with your team if this would be an issue when setting up the CP programme, and were informed that it would not cause issue due to code that runs during the file upload that accounts for the lack of the leading zero._

**Acceptance criteria**
- [ ] Allow non zero errors to be parsed eg EGC000035_A1 and EGC000035_A01
- [ ] Update documentation https://ssg-confluence.internal.sanger.ac.uk/x/jqsKBg (table 3.1)
- [ ] Update  GPL-846 Document de facto plate map standard in pdf to improve Sanger LH communication #246

emrojo:
Closes #253 

* Allow unpadded coordinates (A1, A2, etc) in the files


tests/test_general_helpers.py
```diff
@@ -99,6 +100,23 @@ def test_unpad_coordinate_B01010():
     assert unpad_coordinate("B01010") == "B1010"
 
 
+# tests for pad coordinate
+def test_pad_coordinate_A1():
```
pjvv, line 104: Why not put these asserts in one function? It does not require any fixtures and is quite simplistic.

pjvv (approved):

tests/test_general_helpers.py
emrojo: Ok

PR #283, opened 2021-03-15
Bugfix string comparison

emrojo:
Hot fix - not writing pickable wells in DART because of string equality check bug
@pjvv How do we go about updating the release version?


tests/test_file_processing.py
```diff
@@ -1571,7 +1572,7 @@ def test_insert_plates_and_wells_from_docs_into_dart_multiple_new_plates(config)
             FIELD_PLATE_BARCODE: "TC-rna-00000029",
             FIELD_COORDINATE: "A01",
             FIELD_LAB_ID: "AP",
-            FIELD_RESULT: POSITIVE_RESULT_VALUE,
+            FIELD_RESULT: generate_new_object_for_string(POSITIVE_RESULT_VALUE),
```
pjvv, line 1575: I think it would be more readable to use "Positive" than this function - would this achieve the same result?

tests/conftest.py
```diff
@@ -374,3 +374,17 @@ def blacklist_for_centre(config):
...
+
+
+# For checking bug on comparing strings with 'is'
+# (Pdb) sample.get('Result', False) is 'Positive'
+# <stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
+# False
+# (Pdb) sample2.get('Result', False) is 'Positive'
+# <stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
+# True
+def generate_new_object_for_string(original_str):
```
pjvv, line 386: Move the comments into the function docstring.

pjvv (approved):

pjvv:
All you need to do is increment the `.release_version` file with the appropriate number - this is a bug fix so just update the patch.

tests/test_file_processing.py
emrojo: We tried that, but it wasn't enough to trick Python so it does not re-use the same reference to the string 'Positive' from somewhere else. We also tried: 
```
'Posi'+'tive'
```
but it didnt work either, because python was clever enough to replace this string at pre-execution time with the previous reference of 'Positive' in memory, and that passed the tests in the previous version of code.

pjvv (approved):

tests/test_file_processing.py
pjvv: Interesting!

Issue #356, opened 2021-06-08
DPL-034 UAT Action - Setup Cherrypick Test Data - Crawler elements [Heron]


harrietc52:
**User story**
As PSD we would like an easier way to create cherrypicking test data (source plate data) in the UAT environment. The goal is to have an interactive screen in Lighthouse UI where you select the quantity of plates and numbers of fit to pick samples in them and it triggers the generation of a test data setup via Lighthouse and Crawler.
This story tracks the Crawler elements of this process.

**Who are the primary contacts for this story**
Andrew
Phil
Harriet
Eduardo

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] Add Flask to Crawler with simple API
- [x] Create a POST endpoint to trigger generation of test data, passing the run ID in the body
- [x] Cherrypick Test Data Generation is wrapped in a State Machine
- [x] API endpoint updates test run status in the Mongo collection to 'Started'
- On "Started" state: 
- [x] API endpoint fetches plate information from Mongo collection by run id. 
- [x] API endpoint Calls method to generate data for the specified set of plates
- [x] API endpoint updates test run state in the Mongo collection to 'GeneratedPlateData'
- On "GeneratedPlateData" state:
- [x] API endpoint calls method to write to a Lighthouse csv file and put it into the sftp-data directory in the docker container (under a fake lighthouse directory called CPTD for Cherry Picker Test Data)
- [x] API endpoint updates test run state in the Mongo collection to 'CreatedFile'
- On "CreatedFile" state:
- [x] API endpoint then calls the standard crawler method, specifically targeting the CPTD centre, to parse any outstanding file(s) into the databases
- [x] API endpoint updates test run state in the Mongo collection to 'Completed'
- [x] API endpoint updates test run state in the Mongo collection to 'Failed' if any problem
- [x] A list of plate barcodes and description text are added to the Mongo collection for barcode printing
- [x] API endpoint returns a summary JSON object after all the above has happened.  This will either be a summary of what was created, or a list of errors

**Dependencies**
This story is being done in parallel with these stories:
https://github.com/sanger/lighthouse-ui/issues/73
https://github.com/sanger/lighthouse/issues/322

**Additional context**
LucidChart diagram [here](https://lucid.app/lucidchart/invitations/accept/inv_1b8208a2-17e8-498d-a301-bf34b64f3c6d) showing Lighthouse UI mock ups and UAT Action process

sdjmchattie:
[...]

PR #368, opened 2021-06-30
Dpl 034 wrap crawler in flask

sdjmchattie:
Partial completion of #356 

* Wrap the crawler and its scheduled runs in a Flask app.
* Keep the `runner.py` file as a way of running ad-hoc ingests.


Pipfile
```diff
@@ -15,17 +15,21 @@ black = "*"
...
 pyodbc = "~=4.0"
-colorlog = "~=5.0"
-schedule = "~=1.0"
 pysftp = "~=0.2"
+python-dotenv = "~=0.17"
 slackclient = "~=2.5"
-more-itertools = "~=8.6"
-pandas = "~=1.1"
 sqlalchemy = "~=1.3"
-pymysql = "~=1.0"
```
sdjmchattie, line 28: Sorry for the reordering - it was so I could see what we did and didn't have.  I've added `eve`, three `flask` packages and `python-dotenv`.  I've removed `schedule`

.gitignore
```diff
@@ -136,3 +136,5 @@ dmypy.json
 
 # talisman
 .talismanrc
+
+**/.DS_Store
```
pjvv, line 140: This should be in the system's `.gitignore` file instead of every repo: https://stackoverflow.com/a/7335487

Pipfile
pjvv: Great, I prefer ordered lists! Can you please add `gunicorn` and pin it since it will be required to run in production.

README.md
```diff
@@ -75,32 +75,41 @@ Once all the required packages are installed, enter the virtual environment with
 
     pipenv shell
 
-The following runtime flags are available:
+To then run the app, use the command:
+
+    flask run
+
+This will cause the crawler to execute an ingest every 15 minutes, triggered by cron, so at 0, 15, 30 and 45 minutes past the hour.
```
pjvv, line 82: Sorry for letting the PR languish, can you merge in develop which now uses 30 mins.

README.md
```diff
@@ -75,32 +75,41 @@ Once all the required packages are installed, enter the virtual environment with
...
+
+    SCHEDULER_RUN = False
+
+You can also adjust the behaviour of the scheduled ingest using the settings in the same file.
+To run an ingest immediately, whether Flask is running or not, the `runner.py` file can be used with the arguments shown:
 
     python runner.py --help
 
-    usage: runner.py [-h] [--sftp] [--scheduled]
+    usage: runner.py [-h] [--sftp] [--keep-files] [--add-to-dart] [--centre_prefix {ALDP}]
```
pjvv, line 92: After merging rom develop, the `--centre_prefix` is now a list of all the centres.

.flaskenv
```diff
@@ -0,0 +1,12 @@
...
+
+# https://flask.palletsprojects.com/en/1.1.x/cli/#application-discovery
+FLASK_APP=crawler
+
+# https://flask.palletsprojects.com/en/1.1.x/cli/#setting-command-options
+FLASK_RUN_HOST=0.0.0.0
+FLASK_RUN_PORT=8000
+
+# https://flask.palletsprojects.com/en/1.1.x/config/#environment-and-debug-features
+FLASK_ENV=development
```
pjvv, line 12: Can you move the `SETTINGS_MODULE` env variable to this file too please and remove the `.env` files. It's easier to keep all of our public vars in one place.

crawler/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    app.config.from_object(os.environ["SETTINGS_MODULE"])
+
+    # setup logging
+    logging.config.dictConfig(app.config["LOGGING"])
+
+    if app.config.get("SCHEDULER_RUN", False):
+        scheduler.init_app(app)
+        scheduler.start()
+
+    @app.get("/health")
```
pjvv, line 23: For this healthcheck, I think we should check that there is at least 1 job scheduled. I believe the API for APScheduler makes that available.

crawler/main.py
```diff
@@ -122,3 +125,14 @@ def run(sftp: bool, keep_files: bool, add_to_dart: bool, settings_module: str =
         logger.info("=" * 80)
     except Exception as e:
         logger.exception(e)
+
+
+def scheduled_run():
```
pjvv, line 130: Can we move this to a `jobs` module perhaps?

crawler/priority_samples_process.py
```diff
@@ -154,7 +154,7 @@ def extract_sample_id(sample: SampleDoc) -> ModifiedRowValue:
 
     priority_samples_collection = get_mongo_collection(db, COLLECTION_PRIORITY_SAMPLES)
     for sample_id in sample_ids:
-        priority_samples_collection.update({FIELD_SAMPLE_ID: sample_id}, {"$set": {FIELD_PROCESSED: True}})
+        priority_samples_collection.update_one({FIELD_SAMPLE_ID: sample_id}, {"$set": {FIELD_PROCESSED: True}})
```
pjvv, line 157: Good spot, how did you notice it?

mypy.ini
```diff
@@ -40,5 +40,8 @@ ignore_missing_imports = True
 [mypy-numpy]
 ignore_missing_imports = True
 
-[mypy-schedule]
+[mypy-eve]
```
pjvv, line 43: We don't need this anymore.

runner.py
```diff
@@ -44,32 +35,10 @@
...
-        while True:
-            try:
-                schedule.run_pending()
-                time.sleep(1)
-            except Exception as e:
-                logger.error("There was an exception while running the scheduler")
-                logger.exception(e)
-                # We wait 60 seconds so it wont try to check it again during the same minute
-                time.sleep(60)
+    main.run(sftp=args.sftp, keep_files=args.keep_files, add_to_dart=args.add_to_dart, centre_prefix=args.centre_prefix)
```
pjvv, line 44: I thought we were going to try the flask command line functionality instead? That way, we wouldn't need to worry about processing any env vars as the flask engine would do that for us.

pjvv: Although, just wondering how to hook into the processing when the lighthouse service requests a "crawl".

sdjmchattie: As discussed, let's not change this as it's a development only concern.

README.md
```diff
@@ -75,32 +76,41 @@ Once all the required packages are installed, enter the virtual environment with
...
 
 ## Migrations
 
 ### Updating the MLWH `lighthouse_sample` Table
 
-When the crawler process runs nightly it should be updating the MLWH lighthouse_sample table as it goes with records for
-all rows that are inserted into MongoDB. If that MLWH insert process fails you should see a critical exception for the
-file in Lighthouse-UI. This may be after records inserted correctly into MongoDB, and re-running the file will not
-re-attempt the MLWH inserts in that situation.
+When the crawler process runs every 15 minutes it should be updating the MLWH lighthouse_sample table as it goes with records for all rows that are inserted into MongoDB.
```
pjvv, line 109:
```suggestion
When the crawler process runs every 30 minutes it should be updating the MLWH lighthouse_sample table as it goes with records for all rows that are inserted into MongoDB.
```

crawler/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    def health_check():
+        if scheduler.get_job("run_crawler"):
```
pjvv, line 25: The `run_crawler` job ID should probably be extracted as a constant so that if a change in made in the config it will be picked up here too.

crawler/jobs/apscheduler.py
```diff
@@ -0,0 +1,23 @@
...
+from crawler import scheduler
+from crawler.helpers.general_helpers import get_config
+from crawler.main import run
+
+
+logger = logging.getLogger(__name__)
+
+
+def scheduled_run():
+    """Scheduler's job to do a run every 15 minutes."""
```
pjvv, line 13: Fix comment.

pjvv (changes requested):

pjvv (dismissed):

PR #374, opened 2021-07-05
Added index for lh_sample_uuid to the samples collection

andrewsparkes:


crawler/main.py
```diff
@@ -72,6 +73,10 @@ def run(sftp: bool, keep_files: bool, add_to_dart: bool, settings_module: str =
                 logger.debug(f"Creating index '{FIELD_RESULT}' on '{samples_collection.full_name}'")
                 samples_collection.create_index(FIELD_RESULT)
 
+                # Index on lh_sample_uuid column to make it easier for queries joining on the samples from lighthouse
+                logger.debug(f"Creating index '{FIELD_LH_SAMPLE_UUID}' on '{samples_collection.full_name}'")
+                samples_collection.create_index(FIELD_LH_SAMPLE_UUID)
```
pjvv, line 78: Looks good, as long as it achieves the same as what @emr did manually?

andrewsparkes: I can see the index that @emr made on the samples collection in UAT. If this makes a second instance of it somehow we can drop his version.

emrojo: index was with sparse=true so it wouldnt' include documents that don't have this attribute.

andrewsparkes: ok will add, should it be both **sparse** and **unique** do you think?

andrewsparkes:
Tested in UAT and worked (dropped index manually then ran crawler to recreate index). Took a couple minutes to create the index.

PR #377, opened 2021-07-07
DPL-018 Don't sequence control sample with root sample ID prefix “ZZA” [Heron]

### LINKED ISSUE #347: DPL-018 Don't sequence control sample with root sample ID prefix “ZZA” [Heron]
**User story**
>
DPL-018 | As POC of LH Labs (Cristina A) I would like a known quality control sample with root sample ID prefix “ZZA”, not to be sequenced so we don't waste capacity & money as an Institute on controls

Cristina A

**Acceptance criteria**
- [ ] Filter out samples with  root sample ID prefix “ZZA”
- [ ] Check if these control samples are logged as filtered out. If they are, record it. If not, don't.

**Additional context**
> Cristina A wrote (Thursday, 20 May 2021 at 18:03)

> Glasgow lighthouse (LH) have informed us this week that there is a known quality control sample with root sample ID prefix “ZZA”, which should not be sequenced. While we are asking the LH to avoid sending these samples, I think it might be relevant for us to implement an in-house filter to avoid cherry picking and sequencing unnecessarily – we have already sequenced 8 of these samples.

KatyTaylor:
Closes #347 

* Add a new filtered positive rule to ignore control samples with root sample id starting with 'ZZA'
* We already had a very similar rule in place for 'ZZA000' - make this more generic


andrewsparkes (approved):
Looks good to me. UAT with a test file that has some ZZA's and should be good to go.

PR #376, opened 2021-07-06
DPL-034: Improve unit tests structure

sdjmchattie:
Partially closes #356 

Changes proposed in this pull request:

* Tidy up files in the `tests` directory so that related modules are grouped together.

PR #379, opened 2021-07-07
DPL-034: Add cherrypicker test data generator

sdjmchattie:
Closes #376 

* Add a centre dedicated to test data processing.
* Make centres batch process except for the one above.
* Add a new collection to the Mongo database.
* Add in an endpoint for generating test data according to the specs in the Mongo database and run the crawler on that test data via the new centre.
* Add error handling for poorly formatted inputs and other possible error types from IO, etc.
* Add tests for the above.

Sorry for the size of this PR.  With hindsight, I wonder if I could have broken it into smaller chunks, but everything sort of works together and would have left Crawler a bit of a mess if it wasn't all in at once.


crawler/helpers/cherrypicker_test_data.py
```diff
@@ -0,0 +1,154 @@
...
+def create_rna_id(barcode: str, well_coordinate: str) -> str:
+    return "%s_%s" % (barcode, well_coordinate)
+
+
+def create_rna_pcr_id(barcode: str, well_num: int, well_coordinate: str) -> str:
+    return "RNA_PCR-%s%s_%s" % (barcode, str(well_num).zfill(4), well_coordinate)
+
+
+def create_test_timestamp(dt: datetime) -> str:
+    return dt.strftime("%Y-%m-%d %H:%M:%S UTC")
```
andrewsparkes, line 83: one limitation of the script is the date_tested field is set to more or less the same time for every sample, which is not realistic. maybe we can address this in a later enhancement.

crawler/jobs/cherrypicker_test_data.py
```diff
@@ -0,0 +1,154 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TestDataError(Exception):
+    def __init__(self, message):
+        self.message = message
+
+
+def process(run_id: str, config: Config = None) -> List[List[str]]:
```
andrewsparkes, line 54: Is there a check for environment here somewhere? We really must not let this run in production!

crawler/jobs/cherrypicker_test_data.py
```diff
@@ -0,0 +1,154 @@
...
+    """Generates cherrypicker test data for processing by Crawler and then
+    processes it via the usual runner.
+
+    The specification of the plates to be generated should be in Mongo. Each
+    plate will contain an exact number of positive results between 0 and 96 as
+    specified. Up to 100 plates can be generated at a time.
+
```
andrewsparkes, line 61: The run information stored in Mongo also needs to contain the flag set in the UI about whether the data should be created in DART (required for Beckman) or not (for Biosero). This indicates whether the dart flag be used when calling the crawler runner command to process the files (--add-to-dart).
Later it may also contain a flag per set of plates as to whether those plates should be consolidated ones (no negatives added and one plate per file) or unconsolidated (mixture of positives and negatives, multiple plates per file fine).

andrewsparkes: Also I'm not sure the 100 plates max applies any longer, that was in the script when the id generation was limited to 9999 samples per file.

tests/helpers/test_cherrypicker_data_helper.py
```diff
@@ -0,0 +1,320 @@
...
+    assert actual == expected
+
+
+@pytest.mark.parametrize("count", [2, 3])
+def test_generate_baracoda_barcodes_calls_correct_baracoda_endpoint(request_post_mock, count):
+    expected = ["TEST-012345", "TEST-012346", "TEST-012347"]
+    request_post_mock.return_value.json.return_value = {"barcodes_group": {"barcodes": expected}}
+    actual = generate_baracoda_barcodes(count)
+
+    assert request_post_mock.called_with(f"http://uat.baracoda.psd.sanger.ac.uk/barcodes_group/TEST/new?count={count}")
```
andrewsparkes, line 68: again use config for main part of Baracoda URL

tests/helpers/test_cherrypicker_data_helper.py
```diff
@@ -0,0 +1,320 @@
...
+        positive_rows = [row for row in barcode_rows if "Positive" == row[4]]
+        negative_rows = [row for row in barcode_rows if "Negative" == row[4]]
+        assert len(positive_rows) == positives
+        assert len(negative_rows) == 96 - positives
+
+
+@pytest.fixture
+def test_rows_data():
+    return [
+        ["RSID-01", "VPID-01", "RNAID-01", "RNAPCRID-01", "Positive", "Timestamp", "AP"] + [""] * 12,
```
andrewsparkes, line 257: This doesn't look realistic, the ids should contain plate barcodes and the well numbers are padded to 4 digits, and RNA_ID is <plate_barcode>_<well_coordinate> e.g AP-chp-12345678_A01
But maybe ok if you are just using strings for the tests?

tests/helpers/test_cherrypicker_data_helper.py
```diff
@@ -0,0 +1,320 @@
...
+
+    with patch("builtins.open", side_effect=OSError(5, "Unable to write file")):
+        with pytest.raises(OSError):
+            write_plates_file(test_rows_data, output_path, filename)
+
+    assert is_found_in_list("Exception", logger_messages.error)
+    assert is_found_in_list("Unable to write file", logger_messages.error)
+
+
+def test_create_barcode_meta():
```
andrewsparkes, line 309: To clarify, the idea is to create the label printing list in mongo on the crawler side linked to the run identifier, so it can be pulled out on the lighthouse side for label printing?

crawler/jobs/cherrypicker_test_data.py
```diff
@@ -0,0 +1,154 @@
...
+            update_status(collection, run_id, FIELD_STATUS_PREPARING_DATA)
+
+            csv_rows = create_csv_rows(plate_specs, dt, barcodes)
+            plates_path = os.path.join(config.DIR_DOWNLOADED_DATA, TEST_DATA_CENTRE_PREFIX)
+            plates_filename = f"{TEST_DATA_CENTRE_PREFIX}_{dt.strftime('%y%m%d_%H%M%S_%f')}.csv"
+            write_plates_file(csv_rows, plates_path, plates_filename)
+
+            update_status(collection, run_id, FIELD_STATUS_CRAWLING_DATA)
+
+            run_crawler(sftp=False, keep_files=False, add_to_dart=False, centre_prefix=TEST_DATA_CENTRE_PREFIX)
```
andrewsparkes, line 112: add to dart decision flag needs to be passed through from UI, it is needed to be True for Beckman test data setup and false for Biosero. User decides. Or maybe we just always set it True, can discuss.

crawler/jobs/cherrypicker_test_data.py
```diff
@@ -0,0 +1,154 @@
...
+        run_id: str - The ID of the run.  If this is not found in Mongo an
+            exception will be thrown.
+
+    Returns:
+        Metadata about the plates generated, as:
+        [ [ "barcode1", "description1" ], [ "barcode2", "description2" ] ]
+    """
+    logger.info("Begin generating data.")
+
+    if config is None:
```
andrewsparkes, line 72: quite a long method with lots of nested with if try clauses, possible to refactor?
appreciate some of the other crawler code is much worse on this!

andrewsparkes (changes requested):

tests/helpers/test_cherrypicker_data_helper.py
sdjmchattie: Yes that's what's happening.  This method creates the metadata and then it gets added it to the Mongo document in the `jobs` module.  Lighthouse will see that the run is marked as completed and can then pick that up from the document to be able to send the print off to sprint.

tests/helpers/test_cherrypicker_data_helper.py
sdjmchattie: It shouldn't make any difference.  It's only here as something for the mock to return and to check that it got added verbatim the CSV rows later.

crawler/helpers/cherrypicker_test_data.py
```diff
@@ -0,0 +1,154 @@
...
+    "CH2-Cq",
+    "CH3-Target",
+    "CH3-Result",
+    "CH3-Cq",
+    "CH4-Target",
+    "CH4-Result",
+    "CH4-Cq",
+]
+
+PRINT_FILE_HEADERS = ["barcode", "text"]
```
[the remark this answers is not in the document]
sdjmchattie: No so I've removed it.  Thanks for noticing.

crawler/helpers/cherrypicker_test_data.py
```diff
@@ -0,0 +1,154 @@
...
+
+def create_barcodes(num_required: int) -> list:
+    # call Baracoda here and fetch a set of barcodes with the prefix we want
+    logger.info(f"Num barcodes required from Baracoda = {num_required}")
+    list_barcodes = generate_baracoda_barcodes(num_required)
+    return list_barcodes
+
+
+def create_root_sample_id(barcode: str, well_num: int) -> str:
+    return "RSID-%s%s" % (barcode, str(well_num).zfill(4))
```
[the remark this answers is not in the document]
sdjmchattie: Yes the well_num is always between 1 and 96 only.  If it went above 96, the lookup of the well coordinate in this code further down in the file would fail:

```
well_coordinate = WELL_COORDS[well_index]
```

crawler/helpers/cherrypicker_test_data.py
```diff
@@ -0,0 +1,154 @@
...
+PRINT_FILE_HEADERS = ["barcode", "text"]
+BARACODA_PREFIX = "TEST"
+
+
+def flatten(nested_list: list) -> list:
+    return [item for sublist in nested_list for item in sublist]
+
+
+def generate_baracoda_barcodes(num_required: int) -> list:
+    baracoda_url = f"http://uat.baracoda.psd.sanger.ac.uk/barcodes_group/{BARACODA_PREFIX}/new?count={num_required}"
```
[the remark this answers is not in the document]
sdjmchattie: 👍

crawler/jobs/cherrypicker_test_data.py
sdjmchattie: I was thinking the same thing as I was testing it.  It actually got worse during testing because of the edge cases I realised weren't catered for.  I'll do a refactor now.

crawler/blueprints/cherrypicker_test_data.py
```diff
@@ -0,0 +1,63 @@
+import logging
+
+from datetime import datetime, timezone
+from flask import Blueprint, request
+from flask_cors import CORS
+
+from crawler.constants import (
```
pjvv, line 7: If you are using vscode can I suggest settings similar to these to prevent random diffs because imports get sorted etc.:

```
"[python]": {
    "editor.rulers": [120],
    "editor.codeActionsOnSave": {
      "source.organizeImports": true
    },
    "gitlens.codeLens.symbolScopes": ["!Module"],
    "editor.wordBasedSuggestions": false,
    "editor.formatOnSave": true
  },
```

crawler/blueprints/cherrypicker_test_data.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Returns: FlaskResponse: metadata for the generated test data or a list of
+        errors with the corresponding HTTP status code.
+    """
+    logger.info("Generating test data for cherrypicking hardware")
+
+    timestamp = datetime.utcnow().replace(tzinfo=timezone.utc).isoformat()
+
+    try:
+        if (request_json := request.get_json()) is None or (run_id := request_json.get("run_id")) is None:
```
pjvv, line 43: You might want to check that `run_id` is also not an empty string?

crawler/blueprints/cherrypicker_test_data.py
```diff
@@ -0,0 +1,63 @@
...
+        plates_meta = process(run_id)
+        return ok(run_id=run_id, plates=plates_meta, status=FIELD_STATUS_COMPLETED, timestamp=timestamp)
+
+    except Exception as e:
+        if isinstance(e, TestDataError):
+            msg = str(e)
+        else:
+            msg = f"{FLASK_ERROR_UNEXPECTED} ({type(e).__name__})"
+
+        logger.error(msg)
```
pjvv, line 60: I don't think we need `.error()` and `.exception()`.

pjvv (commented):
Part 1 of review.

crawler/blueprints/cherrypicker_test_data.py
sdjmchattie: That is being validated when we later try to turn it into an `ObjectId()` which generates an exception that the ID is not of the correct format.

crawler/blueprints/cherrypicker_test_data.py
sdjmchattie: Thanks.  I was manually sorting them.  The sort logic I've applied is that external imports come first in alphabetical order.  Then project specific imports afterwards, also in alphabetical order.  I'll use the setting in VSCode instead as it's less fuss for me.

crawler/blueprints/cherrypicker_test_data.py
sdjmchattie: This is logic that I took from Lighthouse: https://github.com/sanger/lighthouse/blob/develop/lighthouse/blueprints/plates.py#L93-L94

But I'm happy to change it to just the error

crawler/jobs/cherrypicker_test_data.py
sdjmchattie: I've added a config value that can be set to True in UAT only.  The default is False and I've then overridden with True in each of the dev environments.  UAT is specified on the deployment project so that will need to be updated to set this option correctly there.

sdjmchattie:
I believe all the comments I haven't marked as resolved are open to discussion about what the functionality should do.  As such I've pushed up my other changes and hope to discuss the other comments tomorrow.

crawler/helpers/cherrypicker_test_data.py
sdjmchattie: I'll add a TODO

crawler/jobs/cherrypicker_test_data.py
sdjmchattie: Plate number limited to 200 instead

andrewsparkes (dismissed):

crawler/config/centres.py
```diff
@@ -52,6 +58,7 @@
         "sftp_root_read": "project-heron_alderly-park",
         "biomek_labware_class": BIOMEK_LABWARE_CLASS_KINGFISHER,
         CENTRES_KEY_SKIP_UNCONSOLIDATED_FILES: True,
+        CENTRES_KEY_INCLUDE_IN_BATCH_PROCESS: True,
```
pjvv, line 61: Just thinking, a small refactor might keep this file a little more clean: when doing `.get(<key>,True)` would be equivalent to having a default value in all the centres. Similar for `CENTRES_KEY_SKIP_UNCONSOLIDATED_FILES`. What do you think?

crawler/constants.py
```diff
@@ -62,12 +62,41 @@
...
+# status field values
+FIELD_STATUS_PENDING: Final[str] = "pending"
+FIELD_STATUS_STARTED: Final[str] = "started"
+FIELD_STATUS_PREPARING_DATA: Final[str] = "preparing_data"
+FIELD_STATUS_CRAWLING_DATA: Final[str] = "crawling_data"
+FIELD_STATUS_COMPLETED: Final[str] = "completed"
+FIELD_STATUS_FAILED: Final[str] = "failed"
+
+###
+# test data
```
pjvv, line 85: Maybe a more descriptive comment here?

crawler/constants.py
```diff
@@ -198,3 +227,12 @@
 
 # As per Beckman events detailed in https://ssg-confluence.internal.sanger.ac.uk/display/PSDPUB/Cherrypicking+Events
 PLATE_EVENT_DESTINATION_CREATED: Final[str] = "lh_beckman_cp_destination_created"
+
+
+###
```
pjvv, line 232: Was this left by mistake?

crawler/helpers/cherrypicker_test_data.py
```diff
@@ -0,0 +1,157 @@
...
+
+from crawler.types import Config
+
+logger = logging.getLogger(__name__)
+
+
+def make_coords_list():
+    for column in "ABCDEFGH":
+        for row in range(1, 13):
+            yield "%s%02d" % (column, row)
```
pjvv, line 17: I think f-string are clearer: `f"{column}{row:02d}"`.

crawler/helpers/cherrypicker_test_data.py
```diff
@@ -0,0 +1,157 @@
...
+BARACODA_PREFIX = "TEST"
+
+
+def flatten(nested_list: list) -> list:
+    return [item for sublist in nested_list for item in sublist]
+
+
+def generate_baracoda_barcodes(config: Config, num_required: int) -> list:
+    baracoda_url = f"{config.BARACODA_BASE_URL}/barcodes_group/{BARACODA_PREFIX}/new?count={num_required}"
+    response = requests.post(baracoda_url, data={})
```
pjvv, line 54: Is `data` required?

crawler/helpers/cherrypicker_test_data.py
```diff
@@ -0,0 +1,157 @@
...
+    for column in "ABCDEFGH":
+        for row in range(1, 13):
+            yield "%s%02d" % (column, row)
+
+
+WELL_COORDS = list(make_coords_list())
+PLATE_SIZE = len(WELL_COORDS)
+
+SAMPLES_FILE_HEADERS = [
+    "Root Sample ID",
```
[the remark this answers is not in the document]
pjvv: We don't need all these field for test data? Just the minimum required would be sufficient unless someone said otherwise?

crawler/helpers/cherrypicker_test_data.py
```diff
@@ -0,0 +1,157 @@
...
+    well_coordinate = WELL_COORDS[well_index]
+    well_num = well_index + 1
+    return [
+        create_root_sample_id(barcode, well_num),
+        create_viral_prep_id(barcode, well_num, well_coordinate),
+        create_rna_id(barcode, well_coordinate),
+        create_rna_pcr_id(barcode, well_num, well_coordinate),
+        result,
+        create_test_timestamp(dt),
+        "AP",
```
pjvv, line 97: This centre seems wrong.

crawler/jobs/cherrypicker_test_data.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def extract_plate_specs(plate_specs_string):
+    try:
+        plate_specs: List[List[int]] = json.loads(plate_specs_string)
+    except (TypeError, json.JSONDecodeError):
+        raise TestDataError(TEST_DATA_ERROR_INVALID_PLATE_SPECS)
+
+    num_plates = reduce(lambda a, b: a + b[0], plate_specs, 0)
+    if num_plates < 1 or num_plates > 200:
```
pjvv, line 141: The higher limit should be extracted to config instead to prevent having a code change to increase?

crawler/types.py
```diff
@@ -26,6 +27,15 @@ class Config(ModuleType):
...
     ADD_LAB_ID: bool
     DIR_DOWNLOADED_DATA: str
+    ENABLE_CHERRYPICKER_ENDPOINTS: bool
+
+    # Ingest Behaviour
+    USE_SFTP: bool
+    KEEP_FILES: bool
+    ADD_TO_DART: bool
+
+    # Baracode
```
pjvv, line 37:
```suggestion
    # Baracoda
```

migrations/helpers/update_filtered_positives_helper.py
```diff
@@ -266,4 +266,4 @@ def biomek_labclass_by_centre_name(centres: List[Dict[str, str]]) -> Dict[str, s
     Returns:
         Dict[str, str] -- biomek labware class by centre name
     """
-    return {centre["name"]: centre["biomek_labware_class"] for centre in centres}
+    return {centre["name"]: centre.get("biomek_labware_class", "") for centre in centres}
```
pjvv, line 269: Hmmm I think we should have a default here - probably the Kingfisher DWP.

mypy.ini
```diff
@@ -42,3 +42,9 @@ ignore_missing_imports = True
 
 [mypy-flask_apscheduler]
 ignore_missing_imports = True
+
+[mypy-flask_cors]
```
pjvv, line 46: We don't need this.

tests/conftest.py
```diff
@@ -385,3 +398,17 @@ def generate_new_object_for_string(original_str):
     part2 = original_str[2:]
     new_str = part1 + part2
     return new_str
+
+
+def is_found_in_list(needle: str, haystack: List[str]) -> bool:
```
pjvv, line 403: Is this more appropriate in a general package not exclusively for tests?

tests/helpers/test_cherrypicker_data_helper.py
```diff
@@ -0,0 +1,320 @@
...
+    flatten,
+    generate_baracoda_barcodes,
+    write_plates_file,
+)
+from tests.conftest import is_found_in_list
+
+LoggerMessages = namedtuple("LoggerMessages", ["info", "error"])
+
+
+@pytest.fixture
```
pjvv, line 30: Shared fixtures should be added to `conftest.py`.

crawler/blueprints/cherrypicker_test_data.py
```diff
@@ -0,0 +1,58 @@
...
+
+from crawler.constants import FIELD_STATUS_COMPLETED, FLASK_ERROR_MISSING_PARAMETERS, FLASK_ERROR_UNEXPECTED
+from crawler.helpers.responses import bad_request, internal_server_error, ok
+from crawler.jobs.cherrypicker_test_data import TestDataError, process
+from crawler.types import FlaskResponse
+
+logger = logging.getLogger(__name__)
+
+bp = Blueprint("cherrypicker", __name__)
+CORS(bp)
```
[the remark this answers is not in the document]
sdjmchattie: I'll remove it

crawler/blueprints/cherrypicker_test_data.py
```diff
@@ -0,0 +1,58 @@
...
+        if (request_json := request.get_json()) is None or (run_id := request_json.get("run_id")) is None:
+            msg = (
+                f"{FLASK_ERROR_MISSING_PARAMETERS} - "
+                "Request body should contain a JSON object with a 'run_id' specified."
+            )
+            logger.error(msg)
+            return bad_request(msg, timestamp=timestamp)
+
+        plates_meta = process(run_id)
+        return ok(run_id=run_id, plates=plates_meta, status=FIELD_STATUS_COMPLETED, timestamp=timestamp)
```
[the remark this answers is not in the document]
sdjmchattie: I thought it would be a useful thing to have in the responses.  Yes the client has its own logs, but it makes resolving issues a lot easier if you know what the server providing the response thought the timestamp was.  You can then relate the logs much more quickly.

crawler/config/centres.py
sdjmchattie: I already do impose the default value in the code, but you can't control how a dictionary gets used especially how someone accessing values chooses to define the default value, compared to an object with a getter.  I agree though that if we are going to declare that there is a default value, that we should remove the re-definition of those defaults in this file.  If I were spending endless time on this, I'd probably create a Centre object and make it enforce the defaults from this config via property getters.

tests/helpers/test_cherrypicker_data_helper.py
sdjmchattie: It's not actually shared if you look carefully.  The logger is already imported into the `cherrypicker_test_data` by the time we get to patch it, so patching the source at `logging.logger` doesn't work.  The patch has to be done on the specific module that this test file is testing.

crawler/helpers/cherrypicker_test_data.py
sdjmchattie: I only did it this way because I didn't know how to apply a number formatter with an f-string.  There's still scope for percentages in some scenarios like where you define the string in a constant and insert values into it later, but that doesn't apply here.  I'll change it.

crawler/constants.py
sdjmchattie: Not sure what you mean?  These constants are used

crawler/helpers/cherrypicker_test_data.py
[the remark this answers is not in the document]
andrewsparkes: It's copied from my script where I had every column. Can lose all the CH columns as they are optional and in any case left empty. And yes the names are in constants already somewhere.

crawler/helpers/cherrypicker_test_data.py
andrewsparkes: Not sure about that one, I copied from other code I'd found, into my script, and Stuart's copied that into here.

crawler/constants.py
pjvv: Oh sorry, because of the empty line on 235 I didn't read the following config as related to the Flask endpoints.

crawler/blueprints/cherrypicker_test_data.py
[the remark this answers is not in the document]
pjvv: Cool with me, it sounds like a good idea.

tests/helpers/test_cherrypicker_data_helper.py
pjvv: Gotcha, thanks for the explanation.

pjvv (approved):

PR #388, opened 2021-07-15
Improve dev experience

sdjmchattie:
Helps to get the dev experience a bit smoother.

* Set the port to 8100 instead of 8000 so it doesn't clash with Lighthouse.
* Separate off the database dependencies into a separate docker-compose file so that Crawler can be started independent of the databases which may have been started in the Lighthouse repository instead.


.release-version
```diff
@@ -1 +1 @@
-1.22.0
+1.22.1
```
pjvv, line 1: Hmm I wouldn't classify this a bug fix but it's not a feature either. Probably closer to a feature than a bug.

Dockerfile
```diff
@@ -53,4 +53,4 @@ CMD ["run"]
 
 # https://docs.docker.com/engine/reference/builder/#healthcheck
 HEALTHCHECK --interval=1m --timeout=3s \
-    CMD curl -f http://localhost:8000/health || exit 1
+    CMD curl -f http://localhost:8100/health || exit 1
```
pjvv, line 56: This healthcheck is used in production too and the production port is set in the gunicorn config. The nginx server also reverse proxies to the port set in gunicorn so we would need to update those in the deployment project too.

dependencies/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+version: "3.8"
+services:
+  # A MySQL service as close to production as currently possible
+  # To run it as a standalone container:
+  #   docker run --name mysql-5.7 -e MYSQL_ALLOW_EMPTY_PASSWORD=yes -d -p 27017:3306 mysql:5.7
```
pjvv, line 5:
```suggestion
  #   docker run --name mysql-5.7 -e MYSQL_ALLOW_EMPTY_PASSWORD=yes -d -p 3306:3306 mysql:5.7
```

dependencies/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+  #   docker run -e 'ACCEPT_EULA=Y' -e 'SA_PASSWORD=MyS3cr3tPassw0rd' -p 1433:1433 -d --name mssql mcr.microsoft.com/mssql/server:2017-latest
+  mssql-2017:
+    image: mcr.microsoft.com/mssql/server:2017-latest
+    ports:
+      - "1433:1433"
+    environment:
+      ACCEPT_EULA: "Y"
+      SA_PASSWORD: MyS3cr3tPassw0rd
+
+volumes:
```
pjvv, line 36: This might need more documentation. If you have a look at the volumes created, when using compose, they are namespaced using the directory they are in. This would lead to confusing results for the inexperienced as data created in the mysql volume for lighthouse would not be available in the mysql volume created by the compose file in crawler.

Dockerfile
sdjmchattie: I'll look at that in the deployment repo.

dependencies/docker-compose.yml
sdjmchattie: Yes indeed ... it is because of this that I've moved them into this sub folder.  The project name that Docker Compose used was `dependencies` according to when you reviewed this PR, but I have updated this further to use a `.env` file to give the project the name I really wanted, which is now `heron_databases`.  I've updated the README to both explain how to run the docker-compose command for these but also to explain further that they are the same databases in both Crawler and Lighthouse.  I will do the same in Lighthouse so that everything is consistent.

If you're worried about the repos going out of sync (I am a little bit), it is possible to put these three files into a separate repo and then refer to them as sub-modules in the Crawler and Lighthouse repos.  This way, they will both be referring to the same files and when they get updated, you don't have to remember to update in both Crawler and Lighthouse.  When you next work on Crawler or Lighthouse it'll advise you there are new updates to the sub-module.

Dockerfile
sdjmchattie: Rather than trying to tweak the deployment, I've changed the repo here to only expose 8100 as part of the `docker-compose.yml` file.  As far as I can tell, a completely different `docker-compose` file is used during deployment so there shouldn't be any conflict.  Please advise if that's not the case.  Developers using Docker for Crawler will get the benefit without impacting the deployment to UAT or production.

dependencies/docker-compose.yml
pjvv: Yes I like the idea of a sub module, we've never used it and it might make a few things easier!

pjvv (approved):

PR #392, opened 2021-07-20
DPL-058-01: Versioned API

sdjmchattie:
Closes: #394 

Decided it was best to start with the Crawler API since it's much smaller and lets us focus on the architecture of the solution.  The V1 blueprint is the only one at the moment and gets served at both `https://crawler-url.sanger/cherrypick-test-data` and also at `https://crawler-url.sanger/v1/cherrypick-test-data`.

In practice I will update the URL being used by Lighthouse and then I'll drop the unversioned endpoint.

When adding a `/v2/` API, the common logic for the existing endpoint can go in another module that isn't decorated.  Each of the two blueprints (v1 and v2) can then call that common module to keep things clean.  If the new endpoint is significantly different, then the logic flow can be directed as needed from these blueprints.


crawler/__init__.py
```diff
@@ -22,20 +22,26 @@ def create_app(config_object: str = None) -> Flask:
...
-
     if app.config.get("SCHEDULER_RUN", False):
         scheduler.init_app(app)
         scheduler.start()
 
+    setup_blueprints(app)
+
     @app.get("/health")
-    def health_check():
+    def _():
```
pjvv, line 32: I know the route indicates the purpose but maybe elaborating in the function's docstring since it's now only `_`.

crawler/__init__.py
```diff
@@ -22,20 +22,26 @@ def create_app(config_object: str = None) -> Flask:
...
 
     return app
+
+
+def setup_blueprints(app):
+    if app.config.get("ENABLE_CHERRYPICKER_ENDPOINTS", False):
+        from crawler.blueprints.v1 import cherrypicker_test_data as cptd_v1
+
+        app.register_blueprint(cptd_v1.bp, url_prefix="/v1")
+        app.register_blueprint(cptd_v1.bp)  # Also serve v1 at the root of the host for now
```
pjvv, line 46: I think it's worth making the change in lighthouse now to prevent this TODO kicking around for too long.

dependencies/down.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
pjvv, line 1: See previous comment.

tests/blueprints/v1/test_cherrypicker_data_blueprint.py
```diff
@@ -4,21 +4,30 @@
 
 import pytest
 
-from crawler.constants import FIELD_STATUS_COMPLETED, FLASK_ERROR_MISSING_PARAMETERS, FLASK_ERROR_UNEXPECTED
+from crawler.constants import (
+    FIELD_STATUS_COMPLETED,
+    FLASK_ERROR_MISSING_PARAMETERS,
+    FLASK_ERROR_UNEXPECTED,
+)
```
pjvv, line 11: What's the reason behind this change?

pjvv (commented):
Nice and clean! I understand the reasoning behind starting in crawler but we were more concerned about versioning APIs which we don't consume ourselves - the few in lighthouse are the most pertinent. When we consume our own APIs we can usually make any required changes in the calling services. I prefer the structure from the stackoverflow post (using `routes.py` etc.) in the lighthouse issue, what did you think of that structure?

README.md
```diff
@@ -293,14 +293,11 @@ you start these dependencies here, there's no need to also attempt to do so in
 the Lighthouse repository. They are the same resources in both and the second
 one to be started will show exceptions about ports already being allocated:
 
-    docker-compose -f dependencies/docker-compose.yml up -d
+    ./dependencies/up.sh
```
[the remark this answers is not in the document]
sdjmchattie: You can use the standard commands.  You just need to switch to the `dependencies` directory before you run `docker-compose up -d`.  Using this `.sh` file means you never get an unexpected outcome from running it.  It works out the relative path to the script, goes into that directory to run `docker-compose up -d` for you and then when the script finished, you haven't moved directory in your own shell.  I thought these conveniences might be worth it to avoid having to explain too much in the README and the inevitable confusion when you run the command from the wrong directory and the dependencies don't start.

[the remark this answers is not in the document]
sdjmchattie: Oh by the way I was going to duplicate these scripts in the lighthouse repo so they both work the same way.  I just can't modify both repos in the same PR 😄

tests/blueprints/v1/test_cherrypicker_data_blueprint.py
sdjmchattie: I didn't explicitly make this change.  It's part of the reformat on save that you said I should add to VSCode.  Even if I  change it back, it goes back to this on save.

pjvv: Hmm something funny is happening, when I save that file it keeps the imports on one line since it does not go over the 120 `line-length` in `pyproject.toml`. Do you have `black` set up as your default formatter for python?

sdjmchattie: I've updated the import sorting parameters in my VSCode like so.  This stops it and `black` from fighting over the style:

```
    "python.sortImports.args": [
        "--multi-line=3",
        "--trailing-comma",
        "--line-length=120"
    ]
```

crawler/__init__.py
sdjmchattie: Will do ... I just need to make that update in another branch then this will be coming out immediately

README.md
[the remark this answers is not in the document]
sdjmchattie: I think we should keep this as shown here unless it turns out to be an issue.  This is the cleanest solution I could come up with for now.

pjvv (dismissed):

tests/routes/v1/test_cptd_endpoints.py
```diff
@@ -30,13 +33,14 @@ def logger_messages():
...
 def process_mock():
-    with patch("crawler.blueprints.cherrypicker_test_data.process") as process:
+    with patch("crawler.routes.common.cherrypicker_test_data.process") as process:
         yield process
 
 
 @pytest.mark.parametrize("json", [{}, {"run_id": None}])
-def test_generate_endpoint_invalid_json(json, client, logger_messages):
-    response = client.post("/cherrypick-test-data", json=json)
+@pytest.mark.parametrize("endpoint_path", ENDPOINT_PATHS)
```
harrietc52, line 41: Does this post to both root and v1 endpoints?

harrietc52 (approved):
Looks good, only thought is the use of "cherrypicker", could it be "cherrypick" to be consistent?

tests/routes/v1/test_cptd_endpoints.py
sdjmchattie: Yes they both get tested.  Of course they should be identical, but I will be removing the root endpoint as soon as I've changed Lighthouse and I'll strip this back to just the v1 endpoint in the test.

sdjmchattie:
> Looks good, only thought is the use of "cherrypicker", could it be "cherrypick" to be consistent?

Thanks @harrietc52 .  I think I'll not rename it at this stage.  It's been `cherrypicker` in the module names since just after I started working on Crawler and not relevant to the change I'm making here.  The endpoints are at least `cherrypick-test-data`.  I didn't know enough about the naming preferences when I started a couple weeks back.

PR #403, opened 2021-08-06
DPL-083: Automatic release versions

sdjmchattie:
Work for https://github.com/sanger/deployment/issues/71

* Only check release versions have been incremented on master (i.e. production builds)
* For develop builds, append the GitHub workflow run ID and `-develop`

Note I chose to use the workflow run ID that increments across all workflows in the repo.  I thought this was better than the ID of the one workflow, as that might get reset if the workflow gets renamed or we have two workflows doing release builds for some reason.


sdjmchattie:
Incidentally, I have no idea if this works or not.  I don't know how I can test it other than to just try to create some releases.

sdjmchattie:
CodeQL is lying.  Those are not "new" alerts.  I don't know if we should be fixing them in a separate PR though.

.github/workflows/check_release_version.yml
```diff
@@ -5,35 +5,23 @@ on:
   pull_request:
     branches:
       - master
-      - develop
```
stevieing, line 8: Just to check. develop is removed from here because we are now automating releases for develop?

.github/workflows/automated_release.yml
```diff
@@ -28,31 +28,33 @@ jobs:
...
           --network host
           --entrypoint ''
-          docker.pkg.github.com/${IMAGE_NAME}:${GITHUB_REF##*/}
+          docker.pkg.github.com/${IMAGE_NAME}:${BRANCH_NAME}
           python -m pytest --no-cov -vx
 
       - name: Set release tag
         # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable
-        run: echo "RELEASE_VERSION=$([ ${GITHUB_REF##*/} = "develop" ] && printf '%s\n' $(cat .release-version)-develop || printf '%s\n' $(cat .release-version))" >> $GITHUB_ENV
+        run: echo "RELEASE_VERSION=$(printf '%s%s\n' $(cat .release-version) $([ ${BRANCH_NAME} = "develop" ] && printf '.%s-develop' ${GITHUB_RUN_ID} || echo ""))" >> $GITHUB_ENV
```
stevieing, line 57: I presume GITHUB_RUN_ID is a variable provided in the action?? Maybe a silly question but just wanted to check.

stevieing (approved):

.github/workflows/automated_release.yml
sdjmchattie: Yes it's provided automatically by GitHub when the workflow is run

.github/workflows/check_release_version.yml
sdjmchattie: Yes we don't want to check whether the `.release-version` file has been incremented by hand as we'll be generating unique version in develop anyway.  But when merging `develop` to `master`, the person should also have made a commit in `develop` changing the release version.

Issue #361, opened 2021-06-15
DPL-040 Mark LH data as uncertain when multiple plate maps are uploaded for a plate and well


rl15:
**User story**
As stakeholder in data quality for Heron (Cristina A) I would like when LH labs upload multiple different copies of a plate map the updated metadata for a unique entry of rna-Id and well are marked uncertain or for dates null   

**Who are the primary contacts for this story**
Cristina A (PAM)
Ian W (PAM)

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] Detect when have update for unique combination of rna_id (plate barcode) and well
- [ ] Update any sample records as inconsistent or for dates null  
- [ ] 

**Questions**
Q. If a later plate map has no values for teh plate map ignore or mark as inconsistent?

Q> In previous discusion discussed if could mark samples as duplicates (no backlog item) - does this requirement move need for this? 

**Dependencies**
This story has a non-blocking relationship with:
DPL-016 Stop parsing the unconsolidated plate map data from MK, AP and GLS on date ? [Heron] #346

**Additional context**

emrojo:
I think what they want is to add a new flag to the sample 'inconsistent_update' in case the sample has been reuploaded with some information change. 
My intention would be to:
- [ ] Add new column to mongodb and to mlwh<lighthouse_sample> table with name 'inconsistent_update_at' with a time stamp of when the sample was 'retried' inconsistently.

idwright:
What we want is to have a single row for each rna_id i.e. plate/well in the lighthouse_sample table.
This will massively improve the speed of our queries

I understand that this may have process implications

andrewsparkes:
Plate map file processing was designed for uploading large files with many tens of thousands of rows, as efficiently and quickly as possible. As such it uses bulk inserts to put the data into the tables. This bulk process doesn't allow for individually checking rows one at a time to see if they already exist in certain key combinations or not, and it doesn't perform any updates when the same key combination is seen more than once. All it gives us is a list of failed inserts due to duplication of primary key (root_sample_id, rna_id, result, and lab id).
Now we've moved to mostly consolidated plates and much smaller single plate files, it might be worth revisiting the whole process of how we insert and update data.
There would have to be very strict agreed rules on what we do when we see the 'same' sample, more than one sample in the 'same' well etc. 
There would also need to be a better user interface to feedback any detected issues with files and samples to the people monitoring the data, and a better interface for them to be able to fix and update the data.
[...]

PR #421, opened 2021-09-08
DPL-040: More SQL efficiency

sdjmchattie:
Closes https://github.com/sanger/crawler/issues/361

After speaking to Simon E. again it turns out that the previous optimisation still wasn't very optimal because of nested queries and indexes not being used.  Now I am setting it up to mark all existing rows for the RNA IDs about to be inserted as `is_current = 0`.  Then I am ensuring that the new entries are correctly populated with `is_current = 1` except for where the same RNA ID appears twice in a file.  In that case, all duplicates except the last entry for the RNA ID is marked as `is_current = 0`.

### DIFF
--- a/crawler/file_processing.py
+++ b/crawler/file_processing.py
@@ -1416,13 +1416,13 @@ class CentreFile:
             num {Decimal128} - the number to be tested
 
         Returns:
-            bool - whether the value lies within range
+            bool -- True if all the values provided are not NaN and num is between, or equal to, range_min and range_max
         """
         min_compare = range_min.compare(num.to_decimal())
-        is_within_min = min_compare != Decimal("1") and not min_compare.is_nan()
+        is_within_min = (min_compare != Decimal("1")) and not min_compare.is_nan()
 
         max_compare = range_max.compare(num.to_decimal())
-        is_within_max = max_compare != Decimal("-1") and not max_compare.is_nan()
+        is_within_max = (max_compare != Decimal("-1")) and not max_compare.is_nan()
 
         return is_within_min and is_within_max
 
--- a/crawler/file_processing.py
+++ b/crawler/file_processing.py
@@ -528,7 +528,7 @@ class CentreFile:
             exception (BulkWriteError): Exception with all the failed writes.
         """
         try:
-            wrong_instances = [write_error["op"] for write_error in exception.details["writeErrors"]]
+            wrong_instances = [write_error["op"] for write_error in exception.details["writeErrors"]]  # type: ignore
             samples_collection = get_mongo_collection(self.get_db(), COLLECTION_SAMPLES)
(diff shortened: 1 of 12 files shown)


crawler/file_processing.py
```diff
@@ -1418,7 +1418,13 @@ def is_within_cq_range(range_min: Decimal, range_max: Decimal, num: Decimal128)
         Returns:
             bool - whether the value lies within range
         """
-        return range_min <= num.to_decimal() <= range_max
+        min_compare = range_min.compare(num.to_decimal())
+        is_within_min = min_compare != Decimal("1") and not min_compare.is_nan()
+
+        max_compare = range_max.compare(num.to_decimal())
+        is_within_max = max_compare != Decimal("-1") and not max_compare.is_nan()
```
emrojo, line 1425: I would add brackets () to group the comparison order:
```suggestion
        is_within_max = (max_compare != Decimal("-1")) and (not max_compare.is_nan())
```

crawler/file_processing.py
```diff
@@ -1418,7 +1418,13 @@ def is_within_cq_range(range_min: Decimal, range_max: Decimal, num: Decimal128)
...
+
```
emrojo, line 1426: I would add either a comment (this is still doing this: range_min <= num.to_decimal() <= range_max) or better change the documentation lines of the method with a description of what is the result, as now the code has more logic than it was before.

crawler/helpers/general_helpers.py
```diff
@@ -253,6 +254,31 @@ def map_mongo_sample_to_mysql(doc: SampleDoc, copy_date: bool = False) -> Dict[s
...
+    for sample in reversed(list(samples)):
+        try:
+            rna_id = sample[MLWH_RNA_ID]
+            reversed_samples.append({**sample, MLWH_IS_CURRENT: rna_id not in existing_rna_ids})
+            existing_rna_ids.add(rna_id)
+        except KeyError:
+            # If there is no RNA ID (shouldn't happen) set is_current to False
+            reversed_samples.append({**sample, MLWH_IS_CURRENT: False})
+
+    return list(reversed(reversed_samples))
```
emrojo, line 279: Why do you use reversed?

sdjmchattie: Because we can assume all samples should have `is_current = True` unless there is a duplicate RNA ID later in the file.  We don't know what's later in the file, so it's easier to work from the bottom to the top.  Keep setting `is_current = True` on each entry unless we get to an RNA ID we already processed in this file, in which case set `is_current = False`.  Once we've processed the file from bottom to top, flip it back over again before returning.

emrojo: Ah, now I see; I would add it as a comment then before the loop.

emrojo (approved):

PR #426, opened 2021-09-13
DPl-133 Update the Plate map file names for brants bridge

harrietc52:
Coupled with https://github.com/sanger/sequencescape/pull/3334


emrojo (commented):
Only change to allow underscore too in the file

emrojo (approved):

PR #429, opened 2021-09-16
DPL-040: Priority samples bug fix

### LINKED ISSUE #361: DPL-040 Mark LH data as uncertain when multiple plate maps are uploaded for a plate and well

sdjmchattie:
Closes #361 

* Prevent updates to row from losing the flag for is_current


crawler/db/mysql.py
```diff
@@ -251,8 +239,9 @@ def reset_is_current_flags(cursor: CMySQLCursor, rna_ids: List[str], chunk_size:
 
     total_rows_affected = 0
     for rna_ids_group in rna_ids_groups:
-        rna_ids_sql = format_sql_list_str(rna_ids_group)
-        cursor.execute(SQL_MLWH_MARK_ALL_SAMPLES_NOT_MOST_RECENT % rna_ids_sql)
+        in_array = ",".join(["%s"] * len(rna_ids_group))
+        sql_statement = SQL_MLWH_MARK_ALL_SAMPLES_NOT_MOST_RECENT % ("%s", in_array)
+        cursor.execute(sql_statement, [datetime.now()] + rna_ids_group)
```
sdjmchattie, line 244: Turns out when you have two insertion points in the SQL template, you have to do this massive hoop jumping exercise.  I tried using named insertions, but they only made the situation worse.  This is the only thing I could find that works.  It basically replaces the second argument with a template for insertion, like `%s,%s,%s,%s` for four values.  Then when we execute the SQL, we pass in the IDs we want to look for and they get slotted into our ad-hoc template.

crawler/sql_queries.py
```diff
@@ -80,7 +87,8 @@
 lh_sample_uuid=VALUES(lh_sample_uuid),
 lh_source_plate_uuid=VALUES(lh_source_plate_uuid),
 must_sequence=VALUES(must_sequence),
-preferentially_sequence=VALUES(preferentially_sequence);
+preferentially_sequence=VALUES(preferentially_sequence),
+is_current=VALUES(is_current);
```
sdjmchattie, line 91: The last of the new unit tests fails without this change, proving it's the source of our issue.

tests/db/test_mysql.py
```diff
@@ -148,56 +153,143 @@ def test_partition():
...
+        "Bob's Biotech",
+    )
 
-        cursor.execute("SELECT root_sample_id FROM lighthouse_sample WHERE is_current=1;")
-        rows = [row[0] for row in cursor.fetchall()]
-        assert rows == ["rna_4"]
 
-    finally:
-        cursor.execute("DELETE FROM lighthouse_sample;")
-        connection.close()
```
emrojo, line 203: This block of code was to reset the testing database to where it was before the test. We haven't refactored this to go into a fixture, so this is needed in every test that adds data to the lighthouse_sample table.

sdjmchattie: It is in the fixture.  The cursor gets its connection from another fixture which is also set up to truncate the table.

crawler/db/mysql.py
emrojo: I had similar problems with this, good you found a better way of doing it

emrojo (commented):
Tests dont get reset after running which it can create problems on subsequent re-runs of the tests.

tests/db/test_mysql.py
emrojo: Ok, I see it now in mlwh_rw_db, Thanks

sdjmchattie:
Conversation with @JamesGlover and @emrojo is indicating this is OK to merge, so I'm going to do that and get it deployed.

JamesGlover (approved):

PR #462, opened 2021-12-02
Dpl181 improve crawler

emrojo:
Closes #

*
*
* ...


.github/workflows/ci.yml
```diff
@@ -117,8 +117,11 @@ jobs:
       - name: Setup the test MLWH and Events databases
         run: |
           python setup_test_db.py
+      - name: Start a testing SFTP server
+        run: |
+          docker run -p "3010:22" -v ${PWD}/test:/home/foo -d atmoz/sftp foo:pass:1001
```
emrojo, line 122: Just to clarify, the reason why we start the SFTP service here instead of in the services section of the github actions file, is because we don't have the folder we want to setup in the SFTP until we have downloaded the code from git, and that only happens after all services have started.

crawler/file_processing.py
```diff
@@ -193,9 +194,17 @@ def download_csv_files(self) -> None:
...
             logger.debug(sftp.listdir(sftp_root_read))
 
-            # download all plate map files
+            # download recent plate map files
             logger.info("Downloading plate map files")
-            sftp.get_d(self.centre_config["sftp_root_read"], self.get_download_dir())
+            for csv_file in sftp.listdir(sftp_root_read):
+                sftp.chdir(sftp_root_read)
+                timestamp = sftp.stat(csv_file).st_mtime  # get timestamp of file
+                createtime = datetime.fromtimestamp(timestamp)
```
sdjmchattie, line 202: Is the creation time the one we want?  I would have thought the modified time would have been better.  I don't know much about the creation of the files by the lighthouses, but is there a chance they modify the contents of an existing file and we don't pick up that change?  Or do they always replace files even when they end up using the same filename?

sdjmchattie: It seems like this is actually already the modified time, so it would be good to rename this variable `modified_time`

sdjmchattie (approved):

PR #508, opened 2022-02-10
DPL-201: hook in RabbitMQ

Jonnie-Bevan:
Contains code which

- creates a consumer 
- creates a producer and sends a message, which is received by the consumer.

Both of these use the schema taken from a local RedPanda registry, encode/decode the message accordingly and communicate via RabbitMQ.


sdjmchattie (changes requested):
Code looks good to me.  There are some linting issues that this PR isn't highlighting because they only trigger on merge to `develop` or `master` but I'll show you how to run the linting on your local machine so you can get them tweaked.

schema_poc/producer.py
```diff
@@ -0,0 +1,43 @@
...
+        version = message_and_info["version"]
+        message = message_and_info["message"]
+
+        connection = BlockingConnection(ConnectionParameters("localhost"))
+        channel = connection.channel()
+        channel.queue_declare(queue=queue)
+        channel.basic_publish(
+            exchange=exchange,
+            routing_key=queue,
+            properties=BasicProperties(headers={"subject": subject, "version": version}),
```
sdjmchattie, line 39: `'subject"` and `"version"` are hard coded here, and then referred to in the consumer using values from the schema registry.  Though these happen to be the same strings, they're not the same use case.  We should put these in a separate `constants.py` file or something like that.  They could be called `MESSAGE_PROPERTY_SUBJECT` and `MESSAGE_PROPERTY_VERSION`

schema_poc/schema_registry.py
```diff
@@ -4,6 +4,7 @@
 
 RESPONSE_KEY_VERSION = "version"
 RESPONSE_KEY_SCHEMA = "schema"
+RESPONSE_KEY_SUBJECT = "subject"
```
sdjmchattie, line 7: This isn't actually part of the response from RedPanda.  We should only be using these constants as part of what RedPanda sends us back when we make requests on it.

schema_poc/producer.py
```diff
@@ -0,0 +1,43 @@
...
+        json_writer(string_writer, write_schema, message)
+
+        prepared_message = {"subject": subject, "version": write_schema_version, "message": string_writer.getvalue()}
+
+        return prepared_message
+
+    def send_message(self, message_and_info, exchange, queue):
+        subject = message_and_info["subject"]
+        version = message_and_info["version"]
+        message = message_and_info["message"]
```
sdjmchattie, line 31: Would it be worth the `prepare_message` method turning the properties into a `BasicProperties` object and returning that?  Otherwise you're mixing up the properties with the message body in the return type from that method and then having to parse them all back out to turn them into `BasicProperties` in the end.

sdjmchattie (changes requested):

sdjmchattie (approved):

PR #509, opened 2022-02-10
DPL-201: Update message schemas

### LINKED ISSUE #470: DPL-nnn JSON Schema definition for plate map data message
**User story**
As a PSD developer I want to create a JSON Schema definition in a public repository that defines the format of messages I can receive in the reception queue for plate maps

Eduardo Martin
Andrew Sparkes

**Acceptance criteria**
- [ ] Create a Json Schema defintion of the messages
- [ ] Schema refers to the information of a single row in plate maps currently
- [ ] Schema should include the UUID of the sample as a mandatory field
- [ ] Schema should support all current columns in MLWH lighthouse_sample table (sample info, priorities, cog_uk_id)
- [ ] We have a command that I can run to validate if a message has the valid format
- [ ] We keep this schema definition in a  public shared repository so it can be accessed by other teams (PAM)

We recomend keeping the same column naming that we have now in MongoDB

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

sdjmchattie:
Closes #470 

Updates the create plate map message schema to match the format discussed.

Also adds in a new message schema for updating fields on an existing sample.


Jonnie-Bevan (approved):

schema_poc/create_plate_map_sample_v1.avsc
```diff
@@ -0,0 +1,89 @@
...
+                "doc": "A plate map sample.",
+                "fields": [
+                  {
+                    "name": "sampleUuid", "type": "string",
+                    "doc": "Unique sample ID, assigned by PAM."
+                  },
+                  {
+                    "name": "rootSampleId", "type": "string",
+                    "doc": "Internal ID for the sample from the lighthouse."
+                  },
```
emrojo, line 50: rnaId will be sample name in SS, I will also vote for receiving it even if it is <plateBarcode>_<PlateCoordinate>

emrojo: I would agree completely if it wasnt because we have to create the sample with the rna_id as the sample name; without it we will need to recreate it again when creating the sample in SS by concatenating with '_'  and coordinates could be also be unpadded so we may not generate the rna_id as it was originally.

KatyTaylor: (I haven't been involved in these discussions at all, so feel free to ignore me, just noticed this conversation)
I would vote for keeping the sample name in the same format going forward - the less that changes, the fewer headaches for reporting and data fixes in the future I think.

schema_poc/test_producer.py
```diff
@@ -1,40 +1,76 @@
...
+    "messageUuid": "UUID-789012-23",
+    "messageCreateDateUtc": datetime.utcnow(),
+    "plate": {
+        "labId": "CPTD",
+        "plateBarcode": "BARCODE001",
+        "samples": [
+            {
+                "sampleUuid": "UUID-123456-01",
+                "rootSampleId": "R00T-S4MPL3-01",
+                "plateCoordinate": "A6",
```
andrewsparkes, line 27: @emrojo Do we want A06 or A6 for coordinates? Best to match in the example test.

andrewsparkes (approved):
Add rnaId then good I think

andrewsparkes (approved):

PR #512, opened 2022-02-14
DPL-198 feedback messages

### LINKED ISSUE #473: DPL-198 - Feedback message JSON Schema (S=S; S=S)
**User story**
As a PSD developer I want to define a JSON Schema for the feedback messages sent to the 'process_status' queue so we can share this definition with other groups about the status of the process of a message originally published in the 'samples_reception' queue. This message will be sent for both successfully processed messages and for errors so that PAM can understand which samples have been successfully received by PSD and which have not.

Eduardo M
Andrew S
Stuart M
Jonnie B

**Acceptance criteria**
- [x] A JSON schema must be defined to support the process_status message
- [x] Every message in the 'process_status' queue is a description of the status of the process from the sample message received in the 'sample_reception' queue
- [x] Every message must contain a message_id with the id of the original message that we are describing the status of process (should be defined in the message from the 'sample_reception' queue
- [x] Every message should have a status with values: 'OK' or 'ERR' indicating that the message_id referred has been correctly processed or not. 
- [x] Every message can contain a list of all errors found during the process of the sample
- [x] Every error described should have an error_type (as defined in confluence) and a human readable message, along with other information if is needed

**Dependencies**
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

sdjmchattie:
Closes #473 

Provides schemas and a proof of concept for messages used to feedback on updates to sample data based on robot events and/or requests to insert and update those samples via other message formats.


Jonnie-Bevan (approved):

schema_poc/test_consumer.py
```diff
@@ -16,5 +16,5 @@
 
 # Read from RabbitMQ
 consumer = Consumer(schema_registry)
-subject = getenv("AVRO_TEST_SUBJECT", "plate-map-sample")
+subject = getenv("AVRO_TEST_SUBJECT", "create-plate-map")
```
andrewsparkes, line 19: could move names like this to a config file

andrewsparkes (approved):

emrojo (approved):

Issue #468, opened 2021-12-13
DPL-203 - Think a way of disabling centres so we can change the source of a specific centres to read plate maps data from RabbitMQ instead of SFTP (S=M;C=M)


emrojo:
**User story**
As a PSD developer I want to have a method that I can be use to enable/disable if a specific centre should read the data from SFTP files or from RabbitMQ. 

Suggestions: 
- it could be the centres table in Mongo that is created from the config files
- it could be commenting/removing the centre from the centres collection
- other options?

**Who are the primary contacts for this story**
Eduardo Martin
Andrew Sparkes

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] I can enable or disable specific centres to read either sftp or Rabbitmq so we can transition slowly by moving centres during transition phase 
- [ ] All code changes related Rabbitmq should not affect SFTP process functionality

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

sdjmchattie:
Trying to get an issue we saw in UAT fixed via pull request https://github.com/sanger/crawler/pull/523

KatyTaylor:
Closing, as is in Done column.

PR #519, opened 2022-03-04
DPL-203 Add data source flag

sdjmchattie:
Closes #https://github.com/sanger/crawler/issues/468

* Add a flag to Centre configs
* Ensure the Mongo database is only populated when it doesn't already exist
* Allow configs to be filtered by data source


crawler/config/centres.py
```diff
@@ -239,3 +263,42 @@
...
+        data_source {str}: The data source filter to apply to centre configs, or None to apply no filter.
+
+    Return:
+        List[CentreConf]: A List of CentreConf from MongoDB matching the given data source.
+    """
+    with create_mongo_client(config) as client:
+        db = get_mongo_db(config, client)
+
+        centres_collection_exists = collection_exists(db, COLLECTION_CENTRES)
+        centres_collection = get_mongo_collection(db, COLLECTION_CENTRES)
```
andrewsparkes, line 283: why get the centres collection here? if in line 282 you find the collection doesn't exist, it won't be there, and you then create it below here in lines starting 285.
I may be misunderstanding what's happening here..

crawler/config/centres.py
```diff
@@ -239,3 +263,42 @@
...
+
+        if not centres_collection_exists:
+            # Populate the centres collection from the config values
+            create_index(centres_collection, FIELD_CENTRE_NAME, unique=True)
+            populate_collection(centres_collection, config.CENTRES, FIELD_CENTRE_NAME)  # type: ignore
+
+        # Get the centres collection from MongoDB
```
andrewsparkes, line 290: and how does this way of getting the collection differ from the method above?

andrewsparkes (commented):
Looks good, decent tests.
Just that one part I'm confused about (see comments).

crawler/config/centres.py
sdjmchattie: Right yes, but I need a reference to it for line 287 and 288 inside the `if` block, and then also outside the `if` block again on line 291.  PyMongo creates the collection with this getter if it doesn't exist.

crawler/config/centres.py
sdjmchattie: The getter you commented on above gets a reference to the collection.  The `if` block populates it with the config file if it didn't already exist.  Then this code here asks the collection reference to get all the documents it contains and casts them to `CentreConf`.  This means we always get the `CentreConf` objects the same way, whether we just inserted them into MongoDB or not.

harrietc52 (commented):
Looks good, haven't looked at python for a while so sorry as I don't have lots to say! Nice cleaning up of constants. 
I wasn't quite following the tests checking the index creation, so maybe you could talk me through them?
Rest looks good though :)

crawler/db/mongo.py
```diff
@@ -82,6 +82,33 @@ def get_mongo_collection(database: Database, collection_name: str) -> Collection
...
+def create_index(collection: Collection, key: str, unique: bool = True) -> None:
+    """Create an index for a specified key on a collection.
+
+    Arguments:
+        collection {Collection}: The collection to create an index on.
+        key {str}: The key to create an index for.
+        unique {bool}: Whether the index may only contain unique values.
+    """
+    logger.debug(f"Creating index '{key}' on '{collection.full_name}'")
+    collection.create_index(key, unique=unique)
```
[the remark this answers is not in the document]
sdjmchattie: Yes, but we don't have that use case yet so I decided to use the YAGNI approach for now.

crawler/config/centres.py
```diff
@@ -239,3 +263,42 @@
...
 ]
+
+
+def get_centres_config(config: Config, data_source: str = "") -> List[CentreConf]:
+    """Get the centres config from MongoDB. If MongoDB does not contain any centres config, it will become populated
+    with the values in the app config for centres.
+
+    Arguments:
+        config {Config}: The configuration object for the whole application.
+        data_source {str}: The data source filter to apply to centre configs, or None to apply no filter.
```
emrojo, line 274: if data_source only admits a limited range of values (rabbitmq and sftp), it would be better typed if defined as enum

sdjmchattie: I did do that, but it doesn't serialise to MongoDB cleanly when I use that.

crawler/main.py
```diff
@@ -44,18 +39,12 @@ def run(sftp: bool, keep_files: bool, add_to_dart: bool, settings_module: str =
         logger.info("START")
         logger.info(f"Using settings from {settings_module}")
 
-        centres = config.CENTRES
+        # get or create the centres collection and filter down to only those with an SFTP data source
+        centres = get_centres_config(config, "SFTP")
```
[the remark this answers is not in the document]
sdjmchattie: Yup, I've made that change in another branch that's now merged

crawler/types.py
```diff
@@ -32,6 +32,7 @@ class CentreConf(TypedDict):
     biomek_labware_class: str
     skip_unconsolidated_surveillance_files: bool
     include_in_scheduled_runs: bool
+    data_source: str
```
emrojo, line 35: It could be enum

tests/config/test_centres.py
```diff
@@ -0,0 +1,57 @@
...
+        ("RaBbItMq", "RabbitMQ"),
+    ],
+)
+def test_get_centres_config_with_data_source(test_data_source_centres, requested, actual_data_source):
+    _, config = test_data_source_centres
+
+    actual = get_centres_config(config, requested)
+
+    assert len(actual) == 1
+    assert actual[0][CENTRE_KEY_DATA_SOURCE] == actual_data_source
```
emrojo, line 32: I dont know if in this test, but I think the value could also have a typo like 'rabbitqm' and in that case, maybe exception?

emrojo (commented):
All good, just added some suggestions, but not important

PR #521, opened 2022-03-07
Unify release, tag & image names

KatyTaylor:
update github build script to unify the naming of github releases, github tags and docker images, to make deployments less confusing


.github/workflows/automated_release.yml
```diff
@@ -61,15 +61,18 @@ jobs:
...
+
+      - name: Set release tag
+        run: echo "RELEASE_TAG=v$RELEASE_NAME" >> $GITHUB_ENV
 
       - name: Create release
         uses: ncipollo/release-action@v1.8.8
         with:
-          name: ${{ env.RELEASE_VERSION }}
-          tag: v${{ env.RELEASE_VERSION }}
+          name: ${{ env.RELEASE_NAME }}
```
KatyTaylor, line 74: So I would give the release name a `v` as well as the tag name, just so there are `v`s everywhere and it's unambiguous if people don't know exactly which string they're meant to use in the Ansible command.

crawler/config/defaults.py
```diff
@@ -38,7 +38,7 @@
 ###
 # Baracoda
 ###
-BARACODA_BASE_URL = f"http://{LOCALHOST}:8000"
+BARACODA_BASE_URL = f"http://{LOCALHOST}:7900"
```
KatyTaylor, line 41: why this change?

sdjmchattie: Oh yeah sorry.  It's partly because I have to make this change manually every time I hook it all up locally.  If you look just below the fold here you'll see that we're using the same port for Cherrytrack which is clearly never going to work.

.github/workflows/automated_release.yml
sdjmchattie: I think I'd rather keep consistency with non-Docker repos for now.  Thanks for the suggestion though.  If it still presents a problem we can change it.

sdjmchattie:
I'm going to merge this to see if it functions correctly before applying it to other repos.

sdjmchattie (approved):

PR #523, opened 2022-03-10
DPL-203: Fix issues from uat

sdjmchattie:
Just taking a stab at the issue we saw when pushing the updated Crawler up to UAT.  Honestly I wish I could explain why this is needed, but I'm struggling to find an answer.  I can't see how CENTRES was populated before this change and I wonder whether it was a side-effect of some code I've changed to the new MongoDB reading approach.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,10 +3,12 @@ name: CI
 on:
   push:
     branches:
+      - develop-*
       - develop
       - master
   pull_request:
     branches:
+      - develop-*
       - develop
       - master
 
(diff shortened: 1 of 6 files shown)


harrietc52 (approved):

emrojo (approved):
All good

.github/workflows/ci.yml
```diff
@@ -3,10 +3,12 @@ name: CI
 on:
   push:
     branches:
+      - develop-*
```
harrietc52, line 6: Can this be removed now?

PR #540, opened 2022-03-31
Merging PAM handover prep work into develop including the proof of concept

sdjmchattie:
Closes #517


README.md
```diff
@@ -55,6 +55,20 @@ The following tools are required for development:
...
+  document to bringing up the dependencies and ensure these are available.
+
+### Additional Dependencies
+
+If you intend to test/develop/use the test data generation functionality of
+Crawler at the `/v1/cherrypick-test-data` endpoint, you will also need to be
+running a local instance of [Baracoda](https://github.com/sanger/baracoda).  The
+default port of Baracoda is 8000, but if you need to use a different port,
+change the port specified for Baracoda in the `crawler/config/defaults.py` file
+in this repository.
```
emrojo, line 70: I think I would include the name of the constant too, so they know what to change; and I will remove the 8000 and only say ' if need to use a different port than the default...' in case that is modified because it seems a comment that can get out of date very easily.

README.md
```diff
@@ -281,15 +295,15 @@ If you do not have root access pyodbc will not work if you use brew. Using the
...
-To run the database dependencies used by Crawler and also Lighthouse, there is a
-separate configuration for Docker Compose. This is shared with Lighthouse so if
-you start these dependencies here, there's no need to also attempt to do so in
-the Lighthouse repository. They are the same resources in both and the second
-one to be started will show exceptions about ports already being allocated:
+To run the dependencies used by Crawler and also Lighthouse, there is a separate
+configuration for Docker Compose. This is shared with Lighthouse so if you start
+these dependencies here, there's no need to also attempt to do so in the
+Lighthouse repository. They are the same resources in both and the second one to
+be started will show exceptions about ports already being allocated:
```
emrojo, line 302: This is a possible problem that people could have. Maybe we could mark it as a **Note**? so is not part of the procedure, but something they should know if they have this issue.

sdjmchattie: This text has been here for months and is reciprocated in the other repositories as well, but I'll change it here if you think it'll help.

sdjmchattie:
No more input imminent from reviewers so I'm going to merge this and see how it performs on UAT

schema_poc/cherrypicked_samples_v1.avsc
```diff
@@ -0,0 +1,25 @@
...
+        "items": {
+          "name": "Sample", "type": "record",
+          "fields": [
+            {
+              "name": "sampleUuid",
+              "doc": "Unique sample UUID.",
+              "type": {"name": "version4SampleUuid", "type": "fixed", "size": 36}
+            }
+          ]
+        }
```
emrojo, line 21: What happens if we have to send this message long after it has been cherrypicked (like the next day) because of service downtime? Do we have a way of specifying the time of cherrypicking for a sample?

sdjmchattie: I don't think we do, no.  However schema definitions were part of https://github.com/sanger/crawler/issues/473 which was completed weeks ago so if we'd like to review decisions made there, it has nothing to do with this PR and we should open a new ticket.  The schemas in the repo here will not be used.  The ones in Redpanda will be.

emrojo (commented):
All good

PR #543, opened 2022-04-04
Prepare Crawler for production release

sdjmchattie:
Closes #517


crawler/config/defaults.py
```diff
@@ -1,12 +1,17 @@
...
+from crawler.config.centres import CENTRES as centres_config
 from crawler.config.logging import *
 from crawler.constants import SCHEDULER_JOB_ID_RUN_CRAWLER
 
 # setting here will overwrite those in 'centres.py'
 
+###
+# centres config objects
+###
+CENTRES = centres_config
```
KatyTaylor, line 13: This is a bit weird, you're importing `CENTRES`, renaming to `centres_config`, then back to `CENTRES`? I'm sure there's a reason :)

crawler/file_processing.py
```diff
@@ -409,7 +404,7 @@ def set_state_for_file(self) -> CentreFileState:
 
         # check for this being an unconsolidated samples file where the centre doesn't support those
         elif (
-            centre.get(CENTRE_KEY_SKIP_UNCONSOLIDATED_SURVEILLANCE_FILES, False)
+            self.centre_config.get(CENTRE_KEY_SKIP_UNCONSOLIDATED_SURVEILLANCE_FILES, False)
```
KatyTaylor, line 407: OK am I understanding this correctly? So before your change it seems like some elements were being retrieved from the `centre_config` that was passed into the `Centre` class, while some were being retrieved from the database. After your change, they're all being retrieved from the `centre_config` that's passed in, but it looks to me like the only place this class is instantiated (outside tests) is in `main.py`, where you're passing in the centres from the db as the argument. So it's all actually consistently coming from the db now.

Is that right? If so seems like a really nice change 👍

sdjmchattie: Yes I think what you described is the intended change.  No more looking up the database mid processing.  All entries come from the database and the lookup is one time when the run begins.

crawler/helpers/cherrypicker_test_data.py
```diff
@@ -48,10 +53,33 @@ def flatten(nested_list: list) -> list:
...
+        except requests.ConnectionError as e:
+            retries = retries - 1
+            logger.error(TEST_DATA_ERROR_BARACODA_CONNECTION)
+            except_obj = CherrypickerDataError(f"{TEST_DATA_ERROR_BARACODA_CONNECTION} -- {str(e)}")
+        except Exception:
+            retries = retries - 1
+            logger.error(TEST_DATA_ERROR_BARACODA_UNKNOWN)
+
+    if except_obj is not None:
+        raise except_obj
```
KatyTaylor, line 81: If the second try hits a connection error, and the third-and-final try hits a generic exception, won't this raise the out-of-date connection error, because `except_obj` wasn't reset?

crawler/helpers/exceptions.py
```diff
@@ -0,0 +1,3 @@
+class CherrypickerDataError(Exception):
+    def __init__(self, message):
+        self.message = message
```
KatyTaylor, line 3: There's already a file crawler/exceptions.py - but I expect you're aware and this either needed to be under the helpers folder or that's good practice.

crawler/helpers/cherrypicker_test_data.py
sdjmchattie: Yes I expect so.  I didn't write this so better to ask @emrojo and @Sangeetha-Bheeman .  I just extracted the hard-coded strings.  It may be that the behaviour you described is what we want.  CherrypickerDataErrors are treated a bit more usefully for the user than other types of error.  The message is deemed safe to show the user.

crawler/config/defaults.py
KatyTaylor: OK, if you have to import it here I think I would prefer an 'ignore' message for the linter, rather than this renaming - if that's a possibility?

Sorry I don't quite understand the first bit - what problem is importing CENTRES into this file solving? And why is this solution horrible? Happy to chat in person tomorrow if that's easier.

sdjmchattie: I don't understand the first bit either.  See this comment: https://github.com/sanger/crawler/pull/523#issue-1165551521

All I can say is that the centre config wasn't being imported into defaults.py and I wonder whether it was doing so via some tenuous link that I broke when I fixed the way data is loaded in.  Probably only worked before by some amount of fortune, certainly not a direct import like I've included here.

PR #544, opened 2022-04-04
DPL-202 Subscribe to messages

sdjmchattie:
Part of #469 

Changes proposed in this pull request:

* Set up a background consumer for messages on the CRUD queue.
* Add all credentials required to consume messages.
* Provide a setup script for local development to ensure vhost and queue is installed on the local RabbitMQ instance.
* Update the README with instructions for the above.

* Consumes message continuously from the CRUD queue.  At the moment logging them to the console.  Next update this will be a callback method so that we can process them into the database and create a feedback message.

PR #549, opened 2022-04-08
DPL-202: Display connection errors and add CA certificates to connection

sdjmchattie:
Part of #544 

* Ensure that errors while connecting are actually logged.  (They were empty strings before)
* Add CA certificates to the SSL connection options.

### DIFF
--- a/crawler/rabbit/async_consumer.py
+++ b/crawler/rabbit/async_consumer.py
@@ -1,5 +1,6 @@
 import functools
 import logging
+import os
 import ssl
 
 from pika import ConnectionParameters, PlainCredentials, SelectConnection, SSLOptions
@@ -84,7 +85,8 @@ class AsyncConsumer(object):
             credentials=credentials,
         )
         if self._server_details.uses_ssl:
-            ssl_context = ssl.create_default_context(cafile="ca_cert.pem")
+            cafile = os.getenv("REQUESTS_CA_BUNDLE")
+            ssl_context = ssl.create_default_context(cafile=cafile)
             connection_params.ssl_options = SSLOptions(ssl_context)
 
         return SelectConnection(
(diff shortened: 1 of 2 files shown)


crawler/rabbit/async_consumer.py
```diff
@@ -58,7 +85,9 @@ def connect(self):
             credentials=credentials,
         )
         if self._server_details.uses_ssl:
-            connection_params.ssl_options = SSLOptions(create_default_context())
+            cafile = os.getenv("REQUESTS_CA_BUNDLE")
+            ssl_context = ssl.create_default_context(cafile=cafile)
```
sdjmchattie, line 89: This is the part that gives it the CA certificate file.  If `getenv` above returns `None` because the environment variable is not set, then the SSL context will be a standard one like in the removed line above.

crawler/rabbit/async_consumer.py
```diff
@@ -90,7 +119,7 @@ def on_connection_open_error(self, _unused_connection, err):
         :param pika.SelectConnection _unused_connection: The connection
         :param Exception err: The error
         """
-        LOGGER.error("Connection open failed: %s", err)
+        LOGGER.error("Connection open failed: %s", AsyncConsumer._reap_last_connection_workflow_error(err))
```
sdjmchattie, line 122: I had to do this otherwise the connection failure becomes a blank string.  Not sure why Pika chose this approach.

emrojo (approved):

crawler/rabbit/async_consumer.py
andrewsparkes: Maybe add a comment about this?

sdjmchattie: Fair enough.  I'll do that straight onto `develop` since I've merged this.

Issue #472, opened 2021-12-13
DPL-199 - Reporting of errors from different situations (S=L; C=M)


emrojo:
**User story**
As a PSD developer I want to keep using the same Imports view to display all errors for each file processed when using RabbitMQ the same as it was happening with the SFTP processing

**Who are the primary contacts for this story**
Eduardo Martin
Andrew Sparkes
Stuart McHattie

**Acceptance criteria**
To be considered successful the solution must allow:
- [x] I can display the information about errors while processing samples from RabbitMQ
- [x] Errors are aggregated by plate barcode
- [x] If plate barcode is not present, no record is created in imports
- [x] If plate barcode is created, we record the error
- [x] We write a record when sample information is not right
- [x] We write a record when database connection error
- [x] We record the number of samples inside the import record
- [x] We also record any successfully processed plate barcode (no errors)
- [x] We create a new table in confluence describing the new list of errors for this process
- [x] Display the total number of errors with an example of the Top - 5 (first 5 that we found)

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

Issue #469, opened 2021-12-13
DPL-nnn Process messages from RabbitMQ as a data source 


emrojo:
**User story**
As a PSD developer I want to be able to run crawler as a consumer that can read from a queue and process messages with the same results as what we do with SFTP now.

**Who are the primary contacts for this story**
Eduardo Martin
Andrew Sparkes

**Acceptance criteria**
To be considered successful the solution must allow:
- [ ] Crawler started as a consumer connected to a queue
- [ ] Crawler can read messages
- [ ] A consumer will read a message, process it and write it in MongoDB, MLWH and DART all in the same process as it is happening at now
- [ ] All changes should be consistent across databases
- [ ] Messages are the primary source of truth, but in second place is MongoDB
- [ ] Crawler errors during the process will be emitted as a new message in a different queue

**Dependencies**
This story is blocked by the following dependencies:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

PR #550, opened 2022-04-11
DPL-199: Perform simple message processing from RabbitMQ

sdjmchattie:
Part of #472 and https://github.com/sanger/crawler/issues/469

* Set up a processor for incoming messages.
* Do basic validation of the message headers.
* Respond with feedback messages.  (currently only failing if the message headers seem wrong)


sdjmchattie:
As discussed, this appears to be working according to the testing I did on my local machine.  I want to write some unit tests to ensure stability in the functionality as we move forwards, but I'll do that in a separate PR as I want to get this into UAT sooner rather than later.

crawler/rabbit/async_consumer.py
```diff
@@ -25,16 +25,14 @@ class AsyncConsumer(object):
...
-        :param str host: The AMQP host to connect to.
-        :param int port: The AMQP port to connect with.
-        :param str username: The AMQP username with read access to the queue.
-        :param str password: The AMQP password for the username.
-        :param str password: The AMQP virtual host to consume from.
+        :param RabbitServerDetails server_details: The RabbitMQ server connection details.
         :param str queue: The AMQP queue to consume from.
+        :param func process_message: A function to call with details of any messages consumed from the queue.
+                                     This function will be passed the message headers, message body and a function
+                                     to call to acknowledge that the message was processed.
```
emrojo, line 35: It could be an instance of a class that handles the process, would it simplify also the arguments?

sdjmchattie: Actually this comment is out of date.  I use the return value after calling the function so I'll update the comment.

emrojo (approved):

PR #566, opened 2022-04-28
DPL-199: Complete create message validation

sdjmchattie:
Part of #472 

* Do all remaining validation on the raw message before processing into databases.


crawler/constants.py
```diff
@@ -261,9 +261,22 @@
 RABBITMQ_CREATE_FEEDBACK_ORIGIN_ROOT = "root"
 RABBITMQ_CREATE_FEEDBACK_ORIGIN_SAMPLE = "sample"
 
+RABBITMQ_FIELD_COG_UK_ID = "cogUkId"
```
stevieing, line 264: Advisory: I have probably said this before but seeing lots of constants which look like they are related makes me think that they could be in an object. It's difficult for me to say if that is the best option based on looking at the code but these fields are grouped together and passed around a fair bit. I can see areas in the code which may benefit from this approach with a bit of encapsulation.

crawler/processing/create_plate_validator.py
```diff
@@ -0,0 +1,213 @@
+import logging
```
stevieing, line 1: Nice. Putting all of the validation in a single class makes it easier to understand and hopefully makes it more maintainable.

stevieing (approved):
Just added some thoughts...

crawler/constants.py
sdjmchattie: Good shout.  I was copying the existing pattern here, but I think grouping these keys would be a good idea.  Unfortunately if you meant using objects instead of dictionaries to contain the message contents I'm not sure if that's going to work as the `fastavro` library gives you a plain dict object after decoding.  However I can look at this again.  I wasn't having any luck when I tried it before.

crawler/processing/create_plate_validator.py
sdjmchattie: Another benefit is that the errors associated with the validation can be kept as a class variable.  I was originally storing them on the message, but the message is supposed to be generic to the contents until you know how to decode it which made doing things like counting samples more complicated.

crawler/constants.py
sdjmchattie: I think it would be worth doing this as a refactor PR rather than muddying the waters on this one so I'll have a trial of that in another branch.

emrojo (approved):

tests/helpers/test_sample_data_helpers.py
```diff
@@ -0,0 +1,35 @@
...
+from crawler.helpers.sample_data_helpers import normalise_plate_coordinate
+
+
+@pytest.mark.parametrize(
+    "input, expected",
+    [
+        ("", ""),
+        ("Hello", "Hello"),
+        ("A0", "A0"),  # not a valid column, so no normalisation applied
+        ("I1", "I1"),  # not a valid row, so no normalisation applied
```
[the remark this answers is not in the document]
sdjmchattie: They will be identified, but not before this method gets called.  This is applied to all the values as we are finding duplicates (don't want to think that `A1` and `A01` are different values). Then the validator will reject values like `I1`.  The Avro schema itself isn't checking for this as you'd have to list out almost 192 values that plate coordinate could be in the schema.  We could do that if we wanted to, but the schema would be messy to look through.  Could be worth it.

PR #584, opened 2022-05-24
DPL-202: Export samples into MongoDB

sdjmchattie:
Part of #469 

* Map samples from the create message into the appropriate MongoDB format.
* Insert those while handling the case of duplicate samples already existing in the database.


crawler/main.py
```diff
@@ -44,76 +29,32 @@ def run(sftp: bool, keep_files: bool, add_to_dart: bool, settings_module: str =
...
-                        if not keep_files and centre_instance.is_download_dir_walkable:
-                            centre_instance.clean_up()
+            ensure_mongo_collections_indexed(db)
+
+            if centre_prefix:
+                # We are only interested in processing a single centre
+                centres = list(filter(lambda config: config.get(CENTRE_KEY_PREFIX) == centre_prefix, centres))
+            else:
+                # We should only include centres that are to be batch processed
+                centres = list(filter(lambda config: config.get(CENTRE_KEY_INCLUDE_IN_SCHEDULED_RUNS, True), centres))
```
emrojo, line 39: Do we have this flag setting in another pull request? (in config/defaults)?

crawler/processing/create_plate_exporter.py
```diff
@@ -135,4 +167,80 @@ def _record_source_plate_in_mongo_db(self, session: ClientSession) -> ExportResu
...
+            except BulkWriteError as ex:
+                LOGGER.warning("BulkWriteError: will now establish whether this was because of duplicate samples.")
+
+                duplication_errors = list(
+                    filter(lambda x: x["code"] == 11000, ex.details["writeErrors"])  # type: ignore
+                )
+
+                if len(duplication_errors) == 0:
+                    # There weren't any duplication errors so this is not a problem with the message contents!
+                    raise
```
emrojo, line 192: Is this raising the same exception? How do we know what happened in this case?

emrojo (approved):

crawler/processing/create_plate_exporter.py
sdjmchattie: Yes it is the same exception.  You should never raise a new exception in an `except:` block as you lose the stack trace and other information.  By raising it again here, it gets caught in the outer `except:` block on line 213 so that it will be reported.  If that's not clear, we could duplicate lines `214 through 219` here and not re-raise, but it felt unnecessary to do that.

crawler/main.py
sdjmchattie: This is just old line 97 unindented as I removed it from a block that has been removed.  But yes, this was to support sending UAT test data in.  The centre that handles UAT test data creation shouldn't be looking for files on the SFTP to process on scheduled runs.

crawler/processing/create_plate_exporter.py
emrojo: No, just checking, I didnt know raise without arguments would reuse the same exception. That seems correct to me, to keep the same exception.

sdjmchattie: Ah right yes.  You can only do it in an `except:` block.

crawler/main.py
sdjmchattie: Yes all centres have this key set.  It was something I added in summer last year alongside Philip.

crawler/config/centres.py
```diff
@@ -287,7 +281,7 @@ def get_centres_config(config: Config, data_source: str = "") -> List[CentreConf
         if not centres_collection_exists:
             # Populate the centres collection from the config values
             create_index(centres_collection, FIELD_CENTRE_NAME, unique=True)
-            populate_collection(centres_collection, config.CENTRES, FIELD_CENTRE_NAME)  # type: ignore
+            populate_mongo_collection(centres_collection, config.CENTRES, FIELD_CENTRE_NAME)  # type: ignore
```
emrojo, line 284: If this is creating all new data, why is it doing 'upsert'?

sdjmchattie: That's the existing workflow.  I agree it doesn't have to be upsert but in this case they're equivalent because we check for the collection before performing the operation.

PR #585, opened 2022-05-25
DPL-202: Samples into DART

### LINKED ISSUE #469: DPL-nnn Process messages from RabbitMQ as a data source

sdjmchattie:
Closes #469 

* Export plate and samples to DART.
* Handle exceptions so that they end up in imports.


crawler/processing/create_plate_exporter.py
```diff
@@ -135,4 +175,146 @@ def _record_source_plate_in_mongo_db(self, session: ClientSession) -> ExportResu
...
+            return ExportResult(
+                success=False,
+                create_plate_errors=[
+                    CreatePlateError(
+                        type=ErrorType.ExportingPostFeedback,  # This error will only reach the imports record
+                        origin=RABBITMQ_CREATE_FEEDBACK_ORIGIN_ROOT,
+                        description=error_description,
+                    )
+                ],
+            )
```
emrojo, line 277: I would move this to another separate method, seems the error connection behaviour, so if someone is interested in the detail can go to that method to check how it was implemented

emrojo (approved):
All good

stevieing (approved):

crawler/rabbit/background_consumer.py
```diff
@@ -47,7 +51,9 @@ def _maybe_reconnect(self):
             self._consumer_var = None
 
     def _get_reconnect_delay(self):
-        if self._consumer.was_consuming:
+        if self._consumer.had_transient_error:
+            self._reconnect_delay = 30
+        elif self._consumer.was_consuming:
             self._reconnect_delay = 0
         else:
             self._reconnect_delay += 1
```
emrojo, line 59: Why is this increment of 1?

sdjmchattie: I didn't write most of this, as it was provided by Pika, but from what I can understand, the only time you might be incrementing 1 seconds at a time is upon first connection while it is trying to get connected to Rabbit.  Then after it has connected `was_consuming` is set to `True` so you always just try the connection again immediately and repeatedly if it drops.  I've added the extra condition at the top for when we had a transient error which wasn't because of RabbitMQ dropping the connection, but something downstream.  Then I'm making it always wait 30 seconds at a time until the resources become available again and the head message gets correctly processed.

Issue #471, opened 2021-12-13
DPL-200 - Support UPDATE messages coming from RabbitMQ messages (S=L; C=M)


emrojo:
**User story**
As a PSD developer I want to be able to support updating sample information received from RabbitMQ messages by using the sample UUID and update ```preferentially_sequence``` and ```must_sequence```.

**Who are the primary contacts for this story**
Eduardo Martin
Andrew Sparkes
Stuart McHattie

**Acceptance criteria**

**IMPORTANT** This story criteria requires to be checked with the result of DPL-253.

To be considered successful the solution must allow:
- [x] Before running the update we check the updated_at field in MongoDB is older than the message timestamp field of the message received; if that is the case we can update. 
- [x] If the timestamp received is not newer than the MongoDB updated_at, we reject the message.
- [x] When a sample changes its priorities fields and its plate has already been picked then it is rejected and a message is published to error queue
- [x] All updates should go to all databases: MongoDB and DART and be rolled back if there is an error in any of them (data will be in both database or in none). 
- [x] If we are in the uncommon situation that there is data in one of the databases that wasn't written in the other, then we write a logger.error() so it will be logged in Slack.
- [ ] Any error type identifiers have been recorded in Confluence.

**Required Process**

- Check the stored procedure on DART for a pending state.
    - If not pending, feedback that the plate cannot be updated.
    - If unable to access DART, trigger a retry on the message after 30 seconds.
- If the plate was pending, query Cherrytrack on endpoint `/source-plates/PLATE_BARCODE`.
    - If the plate is found on the endpoint, feedback that the plate cannot be updated.
    - If unable to access Cherrytrack, trigger a retry on the message after 30 seconds.
- If the plate was pending and not found in Cherrytrack, find the sample on Mongo.
    - If unable to query Mongo, trigger a retry on the message after 30 seconds.
- If Mongo updated_at time is before this message timestamp, update Mongo with the field changes.
    - If unable to update Mongo, trigger a retry on the message after 30 seconds.
- Once Mongo is updated, feedback to PAM that the change has been made.
- Try to update DART.
    - If unable to, create a log that the change has been missed in DART.
    - No more feedback to PAM.

**Dependencies**
This story is blocked by the following dependencies:
- DPL-253 as the solution of DPL-253 story will affect the acceptance criteria for this story

References
This story has a non-blocking relationship with:
- #<issue_no.>
- sanger/<repo>#<issue_no.>

**Additional context**

PR #593, opened 2022-06-07
DPL-200: Add a processor and message object for update messages and feedback

sdjmchattie:
Part of #471 

* Add a processor for update messages.
* Add message schemas as Python objects and a parser.
* Give only error-free feedback (for now) after an update message has come in.


emrojo (approved):
Ok, so the update action will go in another pull request

PR #598, opened 2022-06-10
DPL-200: Add update message exporter

### LINKED ISSUE #471: DPL-200 - Support UPDATE messages coming from RabbitMQ messages (S=L; C=M)

sdjmchattie:
Closes #471 

* Add the exporter for update messages.
* Handle database connection issues and log if manual intervention would be required.


crawler/processing/update_sample_exporter.py
```diff
@@ -0,0 +1,250 @@
+import copy
```
stevieing, line 1: Maybe a bit more top level documentation on what is going on?

sdjmchattie: Yes that's entirely fair.  Will do that.

sdjmchattie:
@stevieing I've added documentation.  I also refactored a property away that wasn't really being used.  And I identified a bug around properties that was masked by the functionality still appearing to behave as expected.  I've fixed that as well.

stevieing (approved):

PR #620, opened 2022-07-04
Prepare release 2.4.0

sdjmchattie:


stevieing (approved):

crawler/__init__.py
```diff
@@ -29,6 +29,10 @@ def create_app(config_object: str = None) -> flask.Flask:
     logging.config.dictConfig(app.config["LOGGING"])
 
     if app.config.get("SCHEDULER_RUN", False):
+        # Add debug level logging for APScheduler -- note that logging.py needs to enable DEBUG logging for
+        # apscheduler as well for this to appear in the logs.
+        logging.getLogger("apscheduler").setLevel(logging.DEBUG)
```
andrewsparkes, line 34: Should this be INFO? Not sure I'm quite following what overrides what with these logging levels. in config/logging it's set to INFO, but here to DEBUG

PR #622, opened 2022-07-08
Dpl361 02 setup processors config

emrojo:
Changes to be able to setup processor by config

### DIFF
--- a/crawler/processing/rabbit_message_processor.py
+++ b/crawler/processing/rabbit_message_processor.py
@@ -15,10 +15,12 @@ class RabbitMessageProcessor:
         self._basic_publisher = basic_publisher
         self._config = config
 
+        self.__processors: Optional[Dict[str, Any]] = None
+
     @property
     def _processors(self) -> Dict[str, Any]:
         if self.__processors is None:
-            self.__processors: Optional[Dict[str, Any]] = {
+            self.__processors = {
                 subject: self._build_processor_for_subject(subject) for subject in self._config.PROCESSORS.keys()
             }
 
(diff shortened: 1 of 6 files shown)


crawler/processing/rabbit_message_processor.py
```diff
@@ -16,14 +15,18 @@ def __init__(self, schema_registry, basic_publisher, config):
...
-            RABBITMQ_SUBJECT_UPDATE_SAMPLE: UpdateSampleProcessor(
-                self._schema_registry, self._basic_publisher, self._config
-            ),
-        }
+        self._build_processors()
+
+    def _build_processors(self):
+        self._processors = {}
+        for subject in self._config.PROCESSORS.keys():
+            self._processors[subject] = self._build_processor_for_subject(subject)
```
sdjmchattie, line 23: Probably a style choice, but is there any reason not to make this a lazy loaded property getter instead?

i.e. set `self.__processors` to `None` in the constructor, then implement this as:

```suggestion
    @property
    def _processors(self):
        if self.__processors is None:
            self.__processors = {subject: self._build_processor_for_subject(subject) for subject in self._config.PROCESSORS.keys()}
        
        return self.__processors
```

sdjmchattie (dismissed):
Looks great.  Just one query to maybe stay consistent with other areas of the code base.

crawler/processing/rabbit_message_processor.py
emrojo: yes looks good! I'll add it

sdjmchattie (approved):

PR #623, opened 2022-07-11
DPL-206: Create test data via RabbitMQ

sdjmchattie:

* Make feedback for test data creation go to a different queue from other feedback.
* Add centre config to contain the prefix to use for the feedback routing key.
* Generate create messages for test data instead of a CSV file.
* Push those messages onto the usual CRUD queue.
* Keep polling the feedback queue to identify the messages we sent and check them for errors.
* Give up and report an error if the messages process incorrectly or no feedback is received after a delay.


crawler/processing/cptd_processor.py
```diff
@@ -0,0 +1,152 @@
...
+
+        self.__schema_registry: Optional[SchemaRegistry] = None
+        self.__basic_publisher: Optional[BasicPublisher] = None
+        self.__encoder: Optional[AvroEncoder] = None
+        self.__decoder: Optional[AvroEncoder] = None
+
+    def generate_test_data(self, create_plate_messages: list) -> None:
+        """Send the plate messages to RabbitMQ then poll for the feedback messages indicating they were all processed
+        correctly.  If there are any issues doing this, raises a CherrypickerDataError to populate the API response
+        with.
```
emrojo, line 37: The name of the method sounds different to what it does to me

crawler/processing/cptd_processor.py
```diff
@@ -0,0 +1,152 @@
...
+                error_free = feedback_message.operation_was_error_free.value
+                if not error_free:
+                    LOGGER.info(
+                        f"Cherrypicker test data create message with UUID '{source_uuid}' was not processed without "
+                        "errors."
+                    )
+                    break
+
+                unconfirmed_uuids.remove(source_uuid)
+
```
emrojo, line 147: I would refactor this section of code. continue could create problems in future. The output condition is modified dynamically and there are too many breaking points (continue, break).

crawler/processing/rabbit_message_processor.py
```diff
@@ -1,6 +1,7 @@
 import logging
 from typing import Any, Dict, Optional, cast
 
+import crawler.helpers.general_helpers as general_helpers
```
emrojo, line 4: Why in this case it is importing the module instead of the functions from the module we will use?

crawler/rabbit/basic_getter.py
```diff
@@ -0,0 +1,77 @@
...
+from crawler.types import RabbitServerDetails
+
+LOGGER = logging.getLogger(__name__)
+MESSAGE_LOGGER = logging.getLogger(LOGGER_NAME_RABBIT_MESSAGES)
+
+
+@dataclass
+class FetchedMessage:
+    headers: dict
+    body: bytes
```
emrojo, line 20: I didnt know about this, thanks :)

crawler/rabbit/basic_getter.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class BasicGetter:
```
emrojo, line 23: Is this a candidate for lib-lab-share?

tests/helpers/test_general_helpers.py
```diff
@@ -74,6 +78,107 @@
...
+@pytest.mark.parametrize("given_password, expected_password", [[None, "pass"], ["", "pass"], ["password", "password"]])
+def test_get_sftp_connection(
+    config, given_username, expected_username, given_password, expected_password, sftp_connection_class
+):
+    actual = get_sftp_connection(config, username=given_username, password=given_password)
+
+    assert actual == sftp_connection_class.return_value
+
+    sftp_connection_class.assert_called_once_with(
+        host=config.SFTP_HOST, port=config.SFTP_PORT, username=expected_username, password=expected_password, cnopts=ANY
```
emrojo, line 115: didn't know about ANY  👍

crawler/processing/cptd_processor.py
sdjmchattie: Could you elaborate on the problems you'd expect continue to cause?  I'm happy to refactor by extracting logic into private functions if you think it's needed, but I can't see how this could be done without use of continue and break nor whether it would actually make it simpler to follow.

tests/jobs/test_cherrypicker_data_job.py
```diff
@@ -86,24 +79,35 @@ def mongo_collection(mongo_database):
...
+
+@pytest.fixture(autouse=True)
+def create_barcode_meta_method():
+    with patch("crawler.jobs.cherrypicker_test_data.create_barcode_meta") as method:
+        method.return_value = CREATED_BARCODE_METADATA
+        yield method
+
 
-        yield stack
+@pytest.fixture(autouse=True)
```
emrojo, line 95: I haven't used autouse before. Have we checked that this fixture autouse does not create the fixture for tests outside of this file?

crawler/processing/rabbit_message_processor.py
sdjmchattie: I probably should change this to match other imports of internal code.  I'm not sure if we do have a standard in our code base since we tend to import whole built-in modules, like `import logging` above.

The reason I did it here was because it makes the patch in testing a bit more logical and I knew I needed to patch these.

If you use `from x.y.z import method` then you have to patch `crawler.processing.rabbit_message_processor.method` because it becomes part of the module it was imported into.

If you use `import x.y.z` then you can patch `x.y.z.method` which feels more logical. 

We probably shouldn't write the code to make unit tests more logical though.  Unit tests are a second class citizen when considering the design of the code.

tests/testing_objects.py
```diff
@@ -65,31 +65,65 @@
...
-from crawler.rabbit.messages.update_sample_message import FIELD_VALUE as UPDATE_SAMPLE_VALUE
+from crawler.rabbit.messages.parsers.create_plate_message import FIELD_RESULT as CREATE_PLATE_RESULT
+from crawler.rabbit.messages.parsers.create_plate_message import FIELD_RNA_ID as CREATE_PLATE_RNA_ID
+from crawler.rabbit.messages.parsers.create_plate_message import FIELD_ROOT_SAMPLE_ID as CREATE_PLATE_ROOT_SAMPLE_ID
+from crawler.rabbit.messages.parsers.create_plate_message import FIELD_SAMPLE_UUID as CREATE_PLATE_SAMPLE_UUID
+from crawler.rabbit.messages.parsers.create_plate_message import FIELD_SAMPLES as CREATE_PLATE_SAMPLES
+from crawler.rabbit.messages.parsers.create_plate_message import FIELD_TESTED_DATE as CREATE_PLATE_TESTED_DATE
+from crawler.rabbit.messages.parsers.update_sample_message import (
+    FIELD_MESSAGE_CREATE_DATE as UPDATE_SAMPLE_MESSAGE_CREATE_DATE,
+)
```
emrojo, line 120: Why we rename the fields? The name they had was not right? Or if not, should this alias be defined inside the module we are importing from?

emrojo (commented):
All looks good, some comments left

crawler/processing/cptd_processor.py
emrojo: breaks and continues change the flow of the program. I'm not saying not to use it; I'm only saying that this function has a lot of them and could potentially originate future errors when modifying the code (indentation, inner loops, ...). Thinking about other ways of doing it without continues and breaks, it seems to me a flow where any failing step send you to the beginning of the process again, one possibility could be a pipeline pattern? ```process([step1, step2, step3,...])```

tests/testing_objects.py
sdjmchattie: The names clash in the namespace when we're creating three different types of message for testing.  The module shouldn't need to attempt to have uniquely named exports.  It is its own namespace.  When we import them into a new namespace it is our duty to ensure they don't clash.  As some do clash and need qualifiers, I've applied the qualifiers across the board to clarify when creating the test objects that we are using the correct keys.

crawler/rabbit/basic_getter.py
sdjmchattie: It might be.  I did consider this as I was writing it, but I don't think it's a usual use case to pull just one message from the queue and stop consuming.  We want to here so that we can only read the queue for the minimum amount of time, but usually you'd want to use an AsyncConsumer.

crawler/processing/rabbit_message_processor.py
emrojo: It could make more sense on unit test mocking, but I also think that could originate more problems than it solves. Most of what people learn on programming is from code of others and if we write code that works in the logical way, a new programmer could get the impression that mocking with pytest works differently to how it works in reality.

sdjmchattie: Neither of these approaches is wrong.  It's a stylistic choice.  Pytest does work in two different ways depending on how you decided to do your import.  It's because of the way Python works.  If anything, you probably do expect to be able to patch `x.y.z.method` but when you do, your mock isn't used in the tests.  In almost all our unit tests, you have to mock your dependencies as if they are defined in the class you're doing tests on.  This is because we use the `from x.y.z import method` approach for all our imports of Crawler code.  I will change this one to be consistent, but the way we have to do testing is confusing to someone unfamiliar with how and when Python imports work.

tests/testing_objects.py
emrojo: Ok

crawler/processing/rabbit_message_processor.py
sdjmchattie: You can see how the tests are affected by this change in this commit:

https://github.com/sanger/crawler/commit/5f23519f30c8b449e5711932ebc639e72af36b27

tests/jobs/test_cherrypicker_data_job.py
sdjmchattie: I've tested how the scope works.  If I move this fixture's definition to another test file, leaving it otherwise unchanged, the tests in this file fail.  It is only applicable to the file it appears in.

crawler/processing/cptd_processor.py
sdjmchattie: I've modified how we approach the parsing of the feedback message reducing us down to a single continue for a completely missing message and a break when a message we were expecting contains errors.  Is this more appealing or do you want to work on another solution?

emrojo (approved):

PR #632, opened 2022-07-22
DPL361 - Adapting crawler to use library lab_share_lib

emrojo:
Removing tests and classes that are already in the library

More changes in import

Change in import

More changes

Closes #

*
*
* ...


crawler/__init__.py
```diff
@@ -33,8 +32,11 @@ def create_app(config_object: str = None) -> flask.Flask:
         scheduler.start()
 
     config, _ = get_config(config_object or "")
+
```
KatyTaylor, line 35: In `cherrypicker_test_data`, you cast the result of `get_config` to a Tuple, but you don't here - not sure if that's an important difference or not

KatyTaylor (approved):
Looks like a pretty straightforward swap. Just left one question.

crawler/__init__.py
emrojo: mypy was not complaining with this one because config does not have a type associated in this code

PR #643, opened 2022-08-03
DPL-206: CPTD unit tests

sdjmchattie:
Add unit tests for CPTD data generation handlers.

Aim is to add at least 1.45% lost in https://github.com/sanger/crawler/pull/623


sdjmchattie:
Unit tests are now complete so they've gone into the previous branch which hasn't been reviewed yet

PR #662, opened 2022-09-02
DPL-480: Make sure fitToPick is only true is the sample has a positive result

### LINKED ISSUE #661: DPL-480 [BUG] Negative test samples show fit to pick as true when they should be false
**Describe the bug**
Negative samples generated for test data under Lighthouse UI are recorded as fit to pick = true when they should be false.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to Lighthouse UI, under the UAT Actions page
2. Generate any test data where the plate is not 100% positive results
3. Observe in MongoDB that the generated negative samples are marked as `filtered_positive: true`

**Expected behaviour**
Samples with a negative result should have fit to pick / filtered positive set to false.

**Additional context**

sdjmchattie:
Closes #661 

* Make negative samples generated for test data have `fit_to_pick` set to `False` instead of `True`

### DIFF
--- a/crawler/helpers/cherrypicker_test_data.py
+++ b/crawler/helpers/cherrypicker_test_data.py
@@ -132,7 +132,7 @@ def _create_sample(dt: datetime, index: int, result: str, plate_barcode: str) ->
         plateCoordinate=well_coordinate,
         preferentiallySequence=False,
         mustSequence=False,
-        fitToPick=True,
+        fitToPick=True if result == "positive" else False,
         result=result,
         testedDateUtc=dt,
     )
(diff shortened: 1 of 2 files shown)


crawler/helpers/cherrypicker_test_data.py
```diff
@@ -132,7 +132,7 @@ def _create_sample(dt: datetime, index: int, result: str, plate_barcode: str) ->
         plateCoordinate=well_coordinate,
         preferentiallySequence=False,
         mustSequence=False,
-        fitToPick=True,
+        fitToPick=True if result == "positive" else False,
```
andrewsparkes, line 135: Just to check, I noticed value in Mongo is capitalised i.e. 'Positive' not 'positive'. Does that matter for the comparison here (and below)? MySQL is case insensitive right?

andrewsparkes (approved):

crawler/helpers/cherrypicker_test_data.py
sdjmchattie: Yup, it is lowercase at this point.  It doesn't really matter what case it is in at this point.  It gets normalised as it gets put into MongoDB here: https://github.com/sanger/crawler/blob/dpl-480-fix-negative-sample-data/crawler/processing/create_plate_exporter.py#L244

PR #705, opened 2022-11-30
DPL-580: Back populate UUIDs for plates

### LINKED ISSUE #704: DPL-580: UUIDs need to be added to samples to be picked
**Describe the work to be done**
The lab would like to be able to run early plates through Cherrypicker robots but before they do they want to ensure the data is present in our databases so that the picking will succeed.

Upon inspection of the 4,832 samples that exist on 74 plates that will be picked, 3977 of them have a value in the `lh_sample_uuid` column of MLWH and none have a value in the `lh_source_plate_uuid` column.  None of the existing UUIDs are in MongoDB.  We need both types of UUID to exist in both locations without overwriting existing UUIDs.  New UUIDs should be generated for missing ones.

**RT Ticket Number**
[#766405](https://rt.sanger.ac.uk/Ticket/Display.html?id=766405)

**Stakeholders**
ag25, am64, ss50

**Additional context**
We have a [script for generating UUIDs](https://github.com/sanger/crawler/blob/develop/migrations/back_populate_source_plate_and_sample_uuids.py) where they are missing but it expects the UUIDs to be completely missing to work.  We need to update that script to recognise that UUIDs already exist in MongoDB and only add them where they are missing.  It can then overwrite MLWH values with the ones it found in MongoDB/Generated.

But before we run that, we need to bring the UUIDs from MLWH into MongoDB.  So a new script needs to be written that does that task.

sdjmchattie:
Closes #704 

* Extract common script functions.
* Provide a migration to bring UUIDs back from MLWH into MongoDB.


migrations/back_populate_uuids_plate_barcodes.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+class ExceptionSampleCountsForMongoAndMLWHNotMatching(Exception):
+    pass
+
+
+"""
+Iterates over the list of plate barcodes provided in a CSV file.
+Finds all the samples in MongoDB with that plate barcode and iterates over those.
+Looks for the same sample in the lighthouse_sample table using mongodb_id and checks whether it has an lh_sample_uuid.
```
emrojo, line 44: About this, only an observation, that this is only valid for samples that were created while PSD had control on lighthouse_sample table, because the mongodb_id is not populated currently by GSU.

migrations/back_populate_uuids_plate_barcodes.py
```diff
@@ -0,0 +1,202 @@
...
+            sample_docs = get_mongo_samples_for_source_plate(samples_collection, source_plate_barcode)
+
+            # iterate through samples
+            for sample_doc in sample_docs:
+                # will every sample doc have a plate_barcode and lab id?
+                LOGGER.info(f"Sample in well {sample_doc[FIELD_COORDINATE]}")
+
+                # update sample in Mongo ‘samples’ to set lh_sample_uuid and updated_timestamp
+                try:
+                    query = SQL_MLWH_GET_SAMPLE_FOR_MONGO_ID % {MLWH_MONGODB_ID: sample_doc[FIELD_MONGODB_ID]}
```
emrojo, line 97: What if a sample from mongo doesnt have mongodb_id? This would query with NULL and returtn a lot of samples unrelated. In theory is not possible, but I haven't checked. Maybe we should log it? Reject? or raise error?

migrations/back_populate_uuids_plate_barcodes.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+    try:
+        mongo_sample = samples_collection.find_one_and_update(
+            filter={
+                FIELD_MONGODB_ID: sample_doc[FIELD_MONGODB_ID],
+            },
+            update={
+                "$set": {
+                    FIELD_LH_SAMPLE_UUID: sample_doc[FIELD_LH_SAMPLE_UUID],
+                    FIELD_UUID_UPDATED: True,
```
emrojo, line 139: Is this flag needed? Are we going to use it anywhere else?

migrations/back_populate_uuids_plate_barcodes.py
```diff
@@ -0,0 +1,202 @@
...
+    query_mongo = list(
+        # fetch just the mongo ids
+        samples_collection.find({FIELD_PLATE_BARCODE: {"$in": source_plate_barcodes}}, {FIELD_MONGODB_ID: 1})
+    )
+
+    list_mongo_ids = [str(x[FIELD_MONGODB_ID]) for x in query_mongo]
+
+    if len(list_mongo_ids) == 0:
+        # There are no samples to check so this is a valid list of samples
+        return
```
emrojo, line 190: Why is a valid list of samples? It means that none of the samples for the plate barcodes provided are in mongo, but is that right? Shouldnt all the samples be always in mongo?

emrojo: Shouldnt it raise an exception instead?

emrojo (dismissed):
Some edge cases I was thinking of, but I dont think we'll have them. Apart from that all looks good to me

migrations/back_populate_uuids_plate_barcodes.py
andrewsparkes: I think every row in a mongodb collection has to have an id, it's fundamental to how it works?

andrewsparkes (dismissed):
There's a lot of changes in the PR but I couldn't see anything wrong.

migrations/back_populate_uuids_plate_barcodes.py
sdjmchattie: Yes this is the case.  MongoDB doesn't allow a missing `_id` field.  If a document is inserted without one, one gets generated automatically: https://www.mongodb.com/docs/manual/core/document/#:~:text=In%20MongoDB%2C%20each%20document%20stored%20in%20a%20collection,documents%20inserted%20through%20update%20operations%20with%20upsert%3A%20true.

migrations/back_populate_uuids_plate_barcodes.py
sdjmchattie: To be fair, yes we can have it raise if we prefer.  It is probably a sign that the user gave an incorrect list of barcodes to process.  There's nothing bad going to happen to the data if there are no documents to update, though.  And you'd know nothing got updated because you'd get the summary saying no documents were updated successfully and none failed.  I'll still change this to a raise, since it's more explicit about the problem.

migrations/back_populate_uuids_plate_barcodes.py
sdjmchattie: I'll add that to the description.  It's useful to know for people looking at this in the future.

migrations/back_populate_uuids_plate_barcodes.py
emrojo: Ah, ok, if other samples where we updated the uuid have already this flag then is probably better to use it here as well to use the same convention.

PR #706, opened 2022-11-30
DPL-580: Allow existing UUIDs when backpopulating

### LINKED ISSUE #704: DPL-580: UUIDs need to be added to samples to be picked

sdjmchattie:
Closes #704 

Needs #705 merging first.

* Add a workaround for exceptions about existing Sample UUIDs when there are no Source Plate UUIDs for those samples.
* The existing code will avoid overwriting the existing UUIDs.
* I've tested this on the production data via my local databases.


emrojo (approved):

migrations/back_populate_source_plate_and_sample_uuids.py
```diff
@@ -49,6 +52,9 @@ class ExceptionSampleCountsForMongoAndMLWHNotMatching(Exception):
 Assumptions:
 1. checked source plates do not already have lh_source_plate_uuid or lh_sample_uuid in either
 the mongo 'source_plate' or 'samples' collections, or in the MLWH lighthouse_sample table
+  - Where Mongo does have lh_sample_uuid set but no lh_source_plate_uuid, an exception will be raised, but if you want
```
andrewsparkes, line 55: got a feeling we added sample uuid before we added source plate uuid, but not completely sure

andrewsparkes (approved):

migrations/back_populate_source_plate_and_sample_uuids.py
sdjmchattie: Maybe so.  I wonder if that will be something else we see as we try to get all these UUIDs in place.

It doesn't explain why we have sample UUIDs in the MLWH but not in MongoDB, but this use case is now to handle my other script back populating those and leaving the Mongo samples without a source plate UUID while the sample UUID is there.

PR #757, opened 2023-03-16
726-dpl-639-restrict-values-in-the-schema-of-priority-flags-to-booleans

yoldas:
Closes #

* Use newer version of lab-share-lib to enforce schema


yoldas:
We used boolean instead of null boolean union type as the required JSON formatting with type info to validate while reading looks non-intuitive.

emrojo (dismissed):

Pipfile
```diff
@@ -22,7 +22,7 @@ flask = "~=2.2"
 flask-apscheduler = "~=1.12"
 gunicorn = "~=20.1"
 more-itertools = "~=9.1"
-mysql-connector-python = "~=8.0"
+mysql-connector-python = "==8.0.31"
```
sdjmchattie, line 25: You shouldn't need to pin this here.  I think something you ran locally updated the version being installed in the `Pipfile.lock` which is the actual log of which versions to install.  Having 8.0.31 installed satisfies the `"~=8.0"` in this `Pipfile` so there's no need for the pin.

dependencies/scripts/mongo-init-replica-set.sh
```diff
@@ -3,7 +3,7 @@
 mongo <<EOF
 var config = {
     _id: "heron_rs",
-    members: [{ _id: 0, host: "mongo:27017"}]
+    members: [{ _id: 0, host: "127.0.0.1:27017"}]
```
sdjmchattie, line 6: You'll likely find that changing this stops the Docker stack from initialising the replica set properly.

Pipfile
yoldas: I have unpinned mysql-connector-python in Pipfile.

dependencies/scripts/mongo-init-replica-set.sh
sdjmchattie: Actually no I'm wrong.  The port is exposed to the local host.  It has been working for me in the previous format though so I am not sure why you've needed to change it

yoldas: I am running the services in docker container but tests locally. The hostname mongo cannot be resolved. Changing hostname to loopback seemed harmless as the port is exposed as you said.

stevieing:
Closed as being on hold.

